PR #1, opened 2017-11-05
implement excel importer

lsmith77:
also refactored AbstractImporter to be more generic

fixes https://github.com/FriendsOfSylius/SyliusImportExportPlugin/issues/15


composer.json
```diff
@@ -6,10 +6,15 @@
     "require": {
         "php": "^7.1",
         "sylius/sylius": "^1.0",
-        "portphp/portphp": "^1.1",
-        "portphp/csv": "^1.0"
+        "portphp/portphp": "^1.2"
+    },
+    "suggest": {
```
lsmith77: waiting on @ddeboer to tag new releases

src/Importer/PaymentMethodsImporter.php
```diff
@@ -14,7 +15,22 @@
...
     /** @var array */
-    protected $headerKeys = ['Code', 'Gateway', 'Name', 'Instructions'];
+    private $headerKeys = ['Code', 'Gateway', 'Name', 'Instructions'];
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function assertKeys(Reader $reader)
+    {
+        if (!method_exists($reader, 'getColumnHeaders')) {
```
lsmith77: this is quite inelegant .. not sure if it makes asking @ddeboer for another interface here .. or to check for specific instances of the relevant reader classes .. I guess lets see how this evolves with other formats?

src/Resources/config/services.yml
```diff
@@ -10,6 +10,11 @@ services:
         arguments:
             - 0
 
+    sylius.factory.excel_reader:
+        class: Port\Excel\ExcelReaderFactory
+        arguments:
+            - 1
```
lsmith77: I guess such stuff should be configurable ..

src/DependencyInjection/SyliusImportExportExtension.php
```diff
@@ -19,6 +19,18 @@ public function load(array $configs, ContainerBuilder $container)
         $configuration = new Configuration();
         $this->processConfiguration($configuration, $configs);
 
+        if (!class_exists('Port\Csv\CsvReaderFactory')) {
```
lsmith77: this also feels a bit iffy

lsmith77:
@solutionDrive-Alt

lsmith77:
I am not sure how the model the tests for excel .. since excel is just a matter of different config compared to csv .. I guess behat tests also test the configuration .. so it makes sense to duplicate them but not the spec tests?

src/DependencyInjection/SyliusImportExportExtension.php
lsmith77: https://twitter.com/lsmith/status/927481599624609792

src/Importer/PaymentMethodsImporter.php
solutionDrive-Alt: maybe we could have our own ReaderInterface which is implemented by small wrapper-classes (which wrap the actual reader) and delegate the call to the method defined in our own Interface to the corresponding Method of the wrapped Reader (e.g. PortCsv or some XmlReader). I hope my thoughts are understandable and I don't misunderstand your concerns.

src/Resources/config/services.yml
solutionDrive-Alt: yeah probably.

solutionDrive-Alt:
good question, I think most of the phpspec tests would look almost the same, especially if the value returned by the reader is also an array. But on the other hand it wouldn't hurt that much and we would be on the save side and we don't know for sure that we will always use the two phpport-readers.

src/Importer/PaymentMethodsImporter.php
lsmith77: yeah .. I was pondering the same as an option.

semi-related .. I was also pondering to move this code into the `AbstractImporter` ..

solutionDrive-Alt: yes, centralising would be a good idea, if it is possible.

src/Importer/AbstractImporter.php
```diff
@@ -58,21 +59,35 @@ public function __construct(
...
     }
 
     /**
+     * @param Reader $reader
+     *
+     * @throws ImporterException
+     */
+    protected function assertKeys(Reader $reader)
+    {
+        if (!method_exists($reader, 'getColumnHeaders')) {
```
lsmith77: hopefully we can soon remove this if https://github.com/portphp/portphp/pull/64 gets merged

lsmith77:
adde Behat tests .. but left out spec tests as I feel this is redundant at least for now

composer.json
solutionDrive-Alt: is this a reason for you to keep this PR open, or should we merge it?

lsmith77:
we now have behat tests .. skipped phpspec .. partially because I wasn’t entirely sure how to set them up and also because I really feel this is redundant atm because in our codebase we are taking exactly the same code paths.

lsmith77:
@solutionDrive-Alt I guess its your turn to merge if you approve :)

solutionDrive-Alt:
true, just thought your comment above in composer.json could be a reason for you to keep it open.

Issue #8, opened 2017-11-07
add support for enqueue to support messages queues


solutionDrive-Alt:
_From @lsmith77 on October 28, 2017 19:24_

see also https://github.com/php-enqueue

see for some inspiration https://github.com/Sylake/SyliusConsumerPlugin/tree/master/src

_Copied from original issue: solutionDrive/SyliusImportExportPlugin#5_

lsmith77:
for this we likely will need to decouple the parsing from entity creation.

lsmith77:
also relates to #7

solutionDrive-Alt:
also relates to #30

lsmith77:
https://github.com/FriendsOfSylius/SyliusImportExportPlugin/pull/120

Issue #7, opened 2017-11-07
refactor AbstractImporter


solutionDrive-Alt:
_From @lsmith77 on October 28, 2017 19:22_

https://github.com/solutionDrive/SyliusImportExport/pull/9#discussion_r146049391

```
We could do it differently: instead of two specific importer classes extending the main abstract importer class, we have one `ResourcesImporter` which has injected 3 parameters:
- `CsvReaderFactory`
- `ObjectManager`
- `ResourceProcessor` (or sth like this)

This `ResourceProcessor` (with `ResourceProcessorInterface`) would have factory and repository injected, and do all *update or create* logic inside. Even more - we could pass there the whole `$csvReader` iterator (or use something like `iterator_to_array` to pass array, it doesn't matter at all to be honest 😄 ) and operate on this multiple rows inside, and then we could get rid of `ObjectManager` here. There are few benefits of such an approach.
- we make importer class much more easier - it just reads data from csv file and pass it to processor
- we get rid of this smelly `createOrUpdateObject` function
- if we want, the resource processor could not even operate on doctrine or even database - it just needs to implement specific method (`public function process(iterable $dataRows): void`) and inside it can write to database, queue, redis... anything 🚀 
```
_Copied from original issue: solutionDrive/SyliusImportExportPlugin#4_

solutionDrive-Alt:
I think we will need still at least one ResourceImporter per type (csv/excel/etc). With the possibillity to implement a special importer in cases, where the generic importer will not work...

lsmith77:
basically what we need is the ability to
* parse from any source into a flat data structure
* project that flat data structure to an actual entity data structure and persist this data structure

I think our importer should basically delegate those two steps to different sub-systems. However to me the first step would be to move the second part into a different layer by essentially taking the code from here as the basis:
https://github.com/Sylake/SyliusConsumerPlugin/tree/master/src/Projector

solutionDrive-Alt:
Do you see this as a part of this issue, or #8 or a complete new one?

PR #26, opened 2017-11-08
improve how to add a new importer type doc section

lsmith77:


README.md
```diff
@@ -75,7 +75,8 @@ admin overview panel using the event hook system, ie. `admin/tax-categories/`.
   - Implement the importer
  
     ```php
-    class FooImporter extends AbstractImporter
+    // alternatively extend from AbstractImporter
```
lsmith77: maybe we should also rename `AbstractImporter` to `AbstractPortPhpImporter` ?

solutionDrive-Alt: yes, or we get started with #7  and try to solve it without the AbstractImporter

lsmith77: which we also need for message queue support https://github.com/FriendsOfSylius/SyliusImportExportPlugin/issues/8

solutionDrive-Alt: Then I would say we should leave the name of the AbstractImporter be and focus on #7.

Issue #24, opened 2017-11-07
convert import form to use the form component


solutionDrive-Alt:
_From @lsmith77 on November 2, 2017 16:1_

see https://github.com/solutionDrive/SyliusImportExportPlugin/blob/master/src/Resources/views/Crud/import_form.html.twig

_Copied from original issue: solutionDrive/SyliusImportExportPlugin#25_

PR #29, opened 2017-11-11
implement form component for upload form

lsmith77:
fixes https://github.com/FriendsOfSylius/SyliusImportExportPlugin/issues/24

### DIFF
--- a/src/Form/ImportType.php
+++ b/src/Form/ImportType.php
@@ -1,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
+
+namespace FriendsOfSylius\SyliusImportExportPlugin\Form;
+
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
+use Symfony\Component\Form\Extension\Core\Type\FileType;
+use Symfony\Component\Form\FormBuilderInterface;
+
+class ImportType extends AbstractType
+{
+    public function buildForm(FormBuilderInterface $builder, array $options)
+    {
+        $builder
+            ->add('format', ChoiceType::class, [
+                'choices' => [
+                    'CSV' => 'csv',
+                    'Excel' => 'excel',
+                ],
+                'expanded' => false,
+                'multiple' => false,
+                'required' => true,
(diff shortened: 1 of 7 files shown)


lsmith77:
see also https://sylius-devs.slack.com/archives/C3E4BU0QH/p1510409833000020

lsmith77:
@solutionDrive-Alt see the above slack discussion .. what is your point of view here?

solutionDrive-Alt:
What I like about our current solution is the fact, that the form is automatically added to the admin-views, where a Importer is implemented. I can‘t exactly tell, due to my limited experience with sylius/symfony, whether the suggested solutions are better. I am with you, for the fact, that the import-process won‘t be triggered manually over the admin on a daily basis, so the fact that one suggested solution would cause more clicks, wouldn‘t be a problem for me.

lsmith77:
actually I think for now I will just do a hacky solution using a subrequest to render the form. this keeps the PR a bit simpler.

src/Form/ImportType.php
```diff
@@ -0,0 +1,29 @@
...
+use Symfony\Component\Form\FormBuilderInterface;
+
+class ImportType extends AbstractType
+{
+    public function buildForm(FormBuilderInterface $builder, array $options)
+    {
+        $builder
+            ->add('format', ChoiceType::class, [
+                'choices' => [
+                    'CSV' => 'csv',
```
lsmith77: this should eventually be dynamic depending on the resource (ie. country, payment_methods etc).

src/Resources/views/Crud/import.html.twig
```diff
@@ -0,0 +1,4 @@
+{{ render(controller(
```
lsmith77: this is the slightly hacky solution to be able to use the controller to create the form instance

lsmith77:
@solutionDrive-Alt ready for review

src/Resources/views/Crud/import.html.twig
solutionDrive-Alt: since I don't know a better solution, it's ok for me...:-)

PR #30, opened 2017-11-12
Refactor abstract importer

solutionDrive-Alt:
Introduced a ResourceImporter and ResourceProcessor. Integrated with new ImporterResult Object from #31


src/Processor/ResourceProcessor.php
```diff
@@ -0,0 +1,53 @@
...
+    private $headerKeys;
+
+    public function __construct(
+        FactoryInterface $resourceFactory,
+        RepositoryInterface $resourceRepository,
+        array $headerKeys = []
+    ) {
+        $this->resourceFactory = $resourceFactory;
+        $this->resourceRepository = $resourceRepository;
+        $this->headerKeys = $headerKeys;
```
solutionDrive-Alt: Not quite sure where to validate the headerKeys now, because we only get a array into the Processor.

solutionDrive-Alt: But maybe I can check at least that the injected header-Keys are present in the given array.

solutionDrive-Alt:
solves #7

solutionDrive-Alt:
needs additional work to integrate with #31

src/Importer/ResourceImporter.php
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    /**
+     * @param Reader $reader
+     *
+     * @throws ImporterException
+     */
+    protected function assertMethods(Reader $reader)
+    {
+        if (!method_exists($reader, 'getColumnHeaders')) {
```
solutionDrive-Alt: Maybe this is not needed anymore, since the method getColumnHeaders is never called directly besides in specs.

src/Processor/ResourceProcessor.php
```diff
@@ -0,0 +1,65 @@
...
+            $method = 'set' . ucfirst($headerKey);
+            if (!method_exists($resource, $method)) {
+                throw new ImporterException(sprintf('Method for %s not found in resource %s', $method, $resource));
+            }
+            call_user_func_array([$resource, $method], [$data[$headerKey]]);
+        }
+        $this->resourceRepository->add($resource);
+    }
+
+    private function validateMetadata(array $data): void
```
solutionDrive-Alt: Maybe the Metadata-Validation move to its own class to prevent missing the implementation in special-processors (as I did with the PaymentMethodProcessor... :-) ) 

- [ ] Implement validate Metadata in PaymentMethodProcessor or move it to ValidationClass

src/Importer/ResourceImporter.php
solutionDrive-Alt: Here also the fact applies that the implementation of this method is missed when implementing a custom Importer, so getting rid of this method would be a benefit

solutionDrive-Alt:
Readme should be updated, too.

src/Importer/ResourceImporter.php
lsmith77: but how do you now validate that the row data structures match the expected data structure?

src/Processor/ResourceProcessor.php
```diff
@@ -0,0 +1,65 @@
...
+    {
+        $this->validateMetadata($data);
+
+        /** @var ResourceInterface $resource */
+        $resource = $this->resourceRepository->findOneBy(['code' => $data['Code']]);
+
+        if (null === $resource) {
+            $resource = $this->resourceFactory->createNew();
+        }
+        foreach ($this->headerKeys as $headerKey) {
```
lsmith77: maybe we should use https://symfony.com/doc/current/components/property_access.html here?

src/Processor/ResourceProcessor.php
lsmith77: ah ok .. you now validate each row so yeah then we don’t need `getColumnHeader()` anymore .. I guess I was concerned over the performance overhead to always check all array keys on every row. but maybe this is premature optimization.

src/Processor/ResourceProcessor.php
solutionDrive-Alt: Yes, looks cool. I will look into it.

solutionDrive-Alt:
@lsmith77 This PR should now be final. I hope it is not too large... :-) I added also some specs for ImporterResult and injected it into the importer, so the specs could be a bit improved for the ResourceImporter.

lsmith77:
nice!

PR #31, opened 2017-11-12
added result meta data

### LINKED ISSUE #9: collect metadata during import
_From @lsmith77 on October 28, 2017 19:24_

like number of items imported, failures, time consumed

_Copied from original issue: solutionDrive/SyliusImportExportPlugin#6_

lsmith77:
fixed #9 

missing testing of returned `ImportResult` instance

### DIFF
--- a/src/Importer/AbstractImporter.php
+++ b/src/Importer/AbstractImporter.php
@@ -6,6 +6,7 @@ namespace FriendsOfSylius\SyliusImportExportPlugin\Importer;
 
 use Doctrine\Common\Persistence\ObjectManager;
 use FriendsOfSylius\SyliusImportExportPlugin\Exception\ImporterException;
+use FriendsOfSylius\SyliusImportExportPlugin\Exception\ItemIncompleteException;
 use Port\Reader;
 use Port\Reader\ReaderFactory;
 use Sylius\Component\Resource\Factory\FactoryInterface;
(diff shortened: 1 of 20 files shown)


src/Importer/AbstractImporter.php
```diff
@@ -55,19 +56,33 @@ public function __construct(
...
     {
         $reader = $this->readerFactory->getReader(new \SplFileObject($fileName));
 
         $this->assertKeys($reader);
 
-        foreach ($reader as $row) {
-            $this->createOrUpdateObject($row);
+        $result = new ImporterResult();
+        $result->start();
+        foreach ($reader as $i => $row) {
```
lsmith77: I guess we should also think about optionally logging skipped/failed rows

src/Importer/AbstractImporter.php
```diff
@@ -55,19 +56,33 @@ public function __construct(
...
+            try {
+                $this->createOrUpdateObject($result, $row);
+                $result->success($i);
+            } catch (ItemIncompleteException $e) {
```
lsmith77: this feels a bit like using exception for control flow .. one option couild also be to allow `createOrUpdateObject` to return  bool instead.

solutionDrive-Alt: with #30 this problem will move to the ResourceProcessor. This processor will get one row each time and should easily tweaked that way, that its process-method returns a bool.

PR #35, opened 2017-11-21
[WIP] add config options for web_ui, batch_size, fail_on_incomplete, stop_on_failure

lsmith77:
missing some additional tests ..
rollback on failure option might also make sense .. though of course in case a batch size is configured, it might not be possible to rollback everything anymore.

### DIFF
--- a/features/ui/importing_tax_categories_from_csv.feature
+++ b/features/ui/importing_tax_categories_from_csv.feature
@@ -9,7 +9,7 @@ Feature: Importing tax categories from csv with the user interface
 
     @ui
     Scenario: Importing defined tax categories
-        When I browse tax categories
+        When I browse all tax categories
         And I import tax category data from "tax_categories.csv" csv file
         Then I should see a notification that the import was successful
         And I should see 2 tax categories in the list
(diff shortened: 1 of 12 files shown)


features/ui/importing_tax_categories_from_csv.feature
```diff
@@ -9,7 +9,7 @@ Feature: Importing tax categories from csv with the user interface
 
     @ui
     Scenario: Importing defined tax categories
-        When I browse tax categories
+        When I browse all tax categories
```
lsmith77: was necessary due to `Step "I browse tax categories" is already defined in Sylius\Behat\Context\Ui\Admin\ManagingTaxCategoriesContext::iWantToBrowseTaxCategories()` :(

lsmith77: of course another possibility would be to switch to using the Sylius context and extend it for our needs .. however this would then bind us to a newer version of Sylius I guess.

solutionDrive-Alt:
@lsmith77 Just saw, that the title is still [WIP]. But I guess it is final and should be merged anyway?

PR #42, opened 2018-01-26
WIP: Added product processor with taxon for import

redGorillaz:
- Added `src/Processor/ProductProcessor.php`
 - Added `products.csv`
 - modified `src/Ressources/config/service.yml`
 - modified `src/Ressources/config/service_csv.yml`

Did not add test for excel format but should be approximately the same
In the `ProductPrecessor.php`, trying to add taxon with import of products but not working.

### DIFF
--- a/src/Processor/ProductProcessor.php
+++ b/src/Processor/ProductProcessor.php
@@ -1,0 +1,129 @@
+<?php
+/**
+ * Created by PhpStorm.
+ * User: apitchen
+ * Date: 17/01/18
+ * Time: 11:41
+ */
+
+namespace FriendsOfSylius\SyliusImportExportPlugin\Processor;
+
+
+use FriendsOfSylius\SyliusImportExportPlugin\Exception\ImporterException;
+use Sylius\Component\Core\Model\ProductInterface;
+use Sylius\Component\Core\Model\TaxonInterface;
+use Sylius\Component\Core\Repository\ProductRepositoryInterface;
+use Sylius\Component\Product\Factory\ProductFactoryInterface;
+use Sylius\Component\Taxonomy\Factory\TaxonFactoryInterface;
+use Sylius\Component\Taxonomy\Repository\TaxonRepositoryInterface;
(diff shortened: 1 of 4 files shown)


lsmith77:
taken from slack:

```
When i try to import product with taxon I have this mistake 
`Expected argument of type "Sylius\Component\Core\Model\TaxonInterface or be null", "string" given`
```

On what line does this error occur? I don’t see where you set the Taxon on the product, so I wonder what is raising this error.

lsmith77:
Also what is in your custom factories? Does it work with the standard factories?

alexispitchen:
I've made a few changes since then. I've got a simple error now 
```
An exception occurred while executing 'INSERT INTO sylius_taxon_translation (name, slug, description, locale, translatable_id) VALUES (?, ?, ?, ?, ?)' with params ["", null, null, "en_US", 9]:                                  
                                                                                                                    
  SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'slug' cannot be null
```
And if you look in `services.yml` I use the custom factory for both product and taxon

solutionDrive-Alt:
Thx for your PR. I will try to have a look at your branch this evening.

solutionDrive-Alt:
@alexispitchen I just checked out your code and tested the Import via the backend. I also got the error you mentioned, but when I checked the database I saw that the products were created. So I hat a look at you Import-File and recognized the blank line at the end of it and checked the other files, where the last line is no blank one. So I think the problem is, that our importer till now is not able to handle files with a blank line at the end. I will open a new Issue for that, when you can confirm that this is the problem and it works.

oallain:
@alexispitchen will you continue this feature?
If no, i'll try to do it ;)

oallain:
Hi,

No response on it, so i'll try to do it ;) I need it to my own project.

vasilvestre:
Should be closed, oallain made it :)

oallain:
Yes, I close it.

Thanks

PR #49, opened 2018-02-23
Redirect to referer instead of building route programmatically

twojtylak:
If you are using the ImportExport Plugin for custom resources e.g. suppliers, the route names usually won't have the prefix `sylius_admin_` . So maybe it's better to redirect to the referer

### DIFF
--- a/src/Controller/ImportDataController.php
+++ b/src/Controller/ImportDataController.php
@@ -84,8 +84,8 @@ final class ImportDataController
         if ($form->isSubmitted() && $form->isValid()) {
             $this->importData($importer, $form);
         }
-
-        return new RedirectResponse($request->headers->get('referer'));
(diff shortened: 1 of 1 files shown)


lsmith77:
is there a pattern for this in Sylius?

twojtylak:
If you are using the ResourceController it is possible to define the redirect route in the _sylius part.

````yaml
_sylius:
     redirect: referer
````
Otherwise I think there is no other option than evaluating the referer and redirecting to this route.
This is the same procedure used in the [ContactController ](https://github.com/Sylius/Sylius/blob/19e78b2b3e71a05efb5526796e25f74091ecd3b3/src/Sylius/Bundle/ShopBundle/Controller/ContactController.php#L106)

lsmith77:
@twojtylak ok fine for me then .. thx for looking into this!

PR #59, opened 2018-03-04
Add exporter for ui

solutionDrive-Alt:
Very basic first implementation of the exporter integration for the ui. Needs polish.

There still some problems, the most ugly is that the AJAX-loader image is always on after pressing the export - button, since I have no RedirectResponse (maybe there is a better way to send the file...)


stefandoorn:
Maybe we can skip AJAX and just call a specific route which will return a content disposition file download? Customer stays on the same page, but still gets a file download. 

Alternative is to generate stuff on the background and make them available as downloads in the UI, but I think that's overkill for now.

stefandoorn:
Regarding the exporting - I think it would for export be better to have a button on the top right (Export) (next to Create) that takes the same filters (but no limit) as the table under it. A dropdown could be done (I've seen that before in Sylius when creating a product) to select the format. Filename can e.g. be automatically generated (or via a service): `{resource}.{format}`. 

I would do the same with import if possible - using a button to route the user to a new URL instead of doing import from the `index`.

lsmith77:
please rebase on master

stefandoorn:
Todo:

- [ ] Specs for `src/Listener/ExportButtonListenerGridListener.php`
- [ ] Behat features

stefandoorn:
Specs for listener added in separate PR to this branch.

src/Exporter/Plugin/ResourcePluginInterface.php
```diff
@@ -6,8 +6,4 @@
 
 interface ResourcePluginInterface extends PluginInterface
```
lsmith77: what is the purpose of this interface now?

src/Exporter/ResourceExporter.php
```diff
@@ -71,6 +97,14 @@ private function writeDataForId(string $id): void
      */
     private function getDataForId(string $id): array
     {
-        return $this->pluginPool->getDataForId($id);
+        $data = $this->pluginPool->getDataForId($id);
+
+        if (null !== $this->transformerPool) {
+            foreach($data as $key => $value) {
```
lsmith77: hmm are we missing a CS checker? a space is missing here ..

src/Exporter/Transformer/Pool.php
```diff
@@ -0,0 +1,50 @@
...
+        if (null === $this->handler) {
+            $this->registerHandlers();
+        }
+
+        return $this->handler->handle($key, $value);
+    }
+
+    private function registerHandlers(): void
+    {
+        foreach($this->generator->getIterator() as $key => $handler) {
```
lsmith77: missing space here as well

lsmith77:
this of course goes much further than just “add exporter for UI” but looks pretty awesome. I read through the code fairly quickly and just spotted one issue with an interface and some CS issues. for the CS issues I guess we might need to add some things to our travis setup to automate this.

I guess thanks to phpstan phpdoc/type hint consistency is now checked quite well. That being said, not necessarily in this PR .. but I think we should consider dropping redundant phpdoc for input/output types ..

awesome stuff .. especially the type converters for the export!

src/Exporter/Plugin/ResourcePluginInterface.php
stefandoorn: I kept it currently for the fact that in case a `resource` source would differ at some moment from `default` source. Maybe it's better for now to remove it and add it later on if needed.

lsmith77: I would remove it then for now

src/Exporter/ResourceExporter.php
stefandoorn: Hm maybe some PHP-CS not going properly?

lsmith77:
@solutionDrive-Alt I would say one last rebase to get the CS fixer activated and then lets merge!

PR #61, opened 2018-03-04
Remove --dev from composer install suggestion

stefandoorn:


lsmith77:
we put that there since we currently don't have a stable release (or any release for that matter)

dunglas:
There are releases now. It should be merged.

lsmith77:
@oallain FYI

PR #78, opened 2018-03-18
[Export] Add spec for export button grid listener

stefandoorn:
Not that easy to test currently (https://github.com/Sylius/Sylius/issues/9271), but I think this does the job for now.


solutionDrive-Alt:
nice!

Issue #80, opened 2018-03-18
Export based on grid filters


stefandoorn:

stefandoorn:
@pamil Could you advice best way to add a generic export controller that can act similar to ResourceController without extending it (else it's tedious for users of this plugin to get each exporter working)? Would like to use the same filters/data fetching as being used on the 'indexAction' in the ResourceController..

Doing something like this doesn't really get it working (WIP) with such URL: `http://webshop.test.local/app_dev.php/admin/export/country/csv?criteria%5Bcode%5D%5Btype%5D=contains&criteria%5Bcode%5D%5Bvalue%5D=nl&criteria%5Benabled%5D=`

```
        // Get metadata config
        // @todo Allow for non-sylius resources
        $metadata = Metadata::fromAliasAndConfiguration('sylius.' . $exporter, $this->container->getParameter('sylius.resources')['sylius.'.$exporter]);
        $configuration = $this->requestConfigurationFactory->create($metadata, $request);
        
        dump($metadata);
        dump($configuration);

        $resources = $this->resourcesCollectionProvider->get($configuration, $repository);

        dump($resources);
```

I have 2 countries (BE & NL), and in this case only expected NL in the results, but both are. I'm probably overlooking something that makes it pass the `criteria` GET param to the resolver..

stefandoorn:
Ok, adding

```yaml
        _sylius:
            paginate: false
            filterable: true
```

to the routing helps, but now it's 0 results :-)

stefandoorn:
Nm, issue with the fact that 'enabled' didn't have a value while filtering. Seems unrelated.

stefandoorn:
Not sure why on the normal grid an empty value for 'enabled' works (it shows one result), but if in the URL in my example it's not returning a result. Might be a bug in resource resolving?

stefandoorn:
To make it more clear for @pamil: 

I'm using this URL on the grid: `http://webshop.test.local/app_dev.php/admin/export/country/csv?criteria%5Bcode%5D%5Btype%5D=contains&criteria%5Bcode%5D%5Bvalue%5D=nl&criteria%5Benabled%5D=`. 

It returns NL only on the grid. In the database I have NL & BE, that's all. This is expected behavior.

[...]

PR #81, opened 2018-03-18
[Export] [WIP] Add exporter for ui grid criteria (first merge #79)

stefandoorn:
Takes criteria from URL GET and passes them to exporter controller, using resource resolving similar to Sylius' ResourceController. Needs tests, features, and more testing.

Also have an issue as mentioned in #80 where the default criteria returns 0 results because the 'enabled' filter seems not be behave properly. It returns data on the grid itself, but in the export the result is zero..

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,7 +222,15 @@ Define your ResourceExporter in services_bar.yml (at the moment only csv is supp
         - ["HeaderKey0", "HeaderKey1" ,"HeaderKey2"]
         - "@sylius.exporters_transformer_pool" # Optional
      tags:
-        - { name: sylius.exporter, type: foo, format: bar }
+        - { name: sylius.exporter, type: app.foo, format: bar }
+```
+
+Note that `app.foo` is the alias as you have named your resource:
+
+```yaml
+sylius_resource:
+    resources:
+        app.foo:
 ```
 
 Define the PluginPool for your ResourceExporter in services.yml
@@ -260,7 +268,7 @@ Define the Countries-Exporter in services_csv.yml
         - ["Id", "Code" ,"Enabled"]
         - "@sylius.exporters_transformer_pool" # Optional
      tags:
-        - { name: sylius.exporter, type: country, format: csv }
+        - { name: sylius.exporter, type: sylius.country, format: csv }
 ```
 
 Define the PluginPool for the Countries-Exporter in services.yml
--- a/README.md
+++ b/README.md
@@ -320,7 +320,7 @@ Optional add the routing:
 
 ```yaml
 app_export_data_country:
-    path: /export/sylius.country/{format}
+    path: /admin/export/sylius.country/{format}
     methods: [GET]
     defaults:
         resource: sylius.country
--- a/README.md
+++ b/README.md
@@ -254,8 +254,23 @@ Define the Plugin for your FooResource in services.yml
           - "@sylius.repository.foo"
           - "@property_accessor"
           - "@doctrine.orm.entity_manager"
-
 ```
+
+In case you want to use the grid filters (in the admin) to filter your output, add to your routing:
+
+```yaml
+app_export_data_foo:
+    path: /export/sylius.resource/{format}
+    methods: [GET]
+    defaults:
+        resource: sylius.foo
+        _controller: sylius.controller.export_data:exportAction
+        _sylius:
+            filterable: true
+            grid: sylius_admin_foo # Name of defined grid here
+```
+
+In case you don't add it, the UI exporters will still function. They will simply load all data of that resource for the export (similar as CLI).
 
 ### A real example
 Define the Countries-Exporter in services_csv.yml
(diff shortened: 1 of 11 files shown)


src/Controller/ExportDataController.php
```diff
@@ -33,49 +54,50 @@ public function __construct(ServiceRegistryInterface $registry)
...
         if (!$this->registry->has($name)) {
             throw new \Exception(sprintf("No exporter found of type '%s' for format '%s'", $exporter, $format));
         }
 
-        /** @var ResourceExporterInterface $service */
-        $service = $this->registry->get($name);
-
         /** @var \Sylius\Component\Resource\Repository\RepositoryInterface $repository */
-        $repository = $this->get('sylius.repository.' . $exporter);
+        $repository = $this->get('sylius.repository.' . $resource);
```
lsmith77: do we have to add some filtering here to prevent security issues?

stefandoorn:
I'll take a look @lsmith77. First got a WIP for exporting data using Grid Filters. The reason is that for the Index page in the admin, a grid is defined. Each grid can also be extended with extra fields. By default this works, but the criteria that get passed via the query parameters form an issue. They are not being applied to the query properly and it's unsafe to do so as @lsmith77 pointed out. I'm not sure yet how to dynamically create export routings, for now it's just defined by adding routes by hand. This way we can use the grid to get the proper data and work with that. It still works also without a grid, but we need to see whether that is safe.

Also would like to have input from the Sylius core team (cc @pamil) whether this way of doing so is the best or whether they see better methods to apply the grid.

src/Controller/ExportDataController.php
stefandoorn: Filtering for a string e.g. + a `has` check before it?

src/Listener/ExportButtonGridListener.php
```diff
@@ -86,14 +100,21 @@ private function createLinks(): array
...
     private function addLink(string $format): void
     {
+        $parameters = [
+            'resource' => $this->resource,
+            'format' => $format,
+        ];
+
+        $currentRequest = $this->requestStack->getCurrentRequest();
+        if ($currentRequest) {
+            $parameters['criteria'] = $currentRequest->query->get('criteria');
```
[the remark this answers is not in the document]
stefandoorn: I'm not sure how to do this, any suggestion @lsmith77? I think it's similar possible in Sylius, to add whatever you want in the `criteria` query. If a grid is active, Sylius filters it against the grid filters. I think it's a good alternative to disable filtering on routes that have no grid defined, therefore always exporting everything from the db instead of a filtered set. WDYT?

src/Resources/config/routing.yml
```diff
@@ -1,11 +1,27 @@
...
+
+## Backup route
 app_export_data:
     path: /export/{resource}/{format}
     methods: [GET]
     defaults:
         _controller: sylius.controller.export_data:exportAction
+        _sylius:
+            paginate: false
+            filterable: true
```
stefandoorn: @lsmith77 @solutionDrive-Alt WDYT about adjusting `filterable` to `false` for the backup export route. If dev wants to add filtering based on grid filters, they should define their own route as shown some lines up. 

I haven't found a solution yet to automate adding routes with the same grid setting as the admin index routes for the same resource (or hook into Sylius admin route generation from the plugin - then I can get it working..). If anyone has an idea for that, I'm open to that, then we can get rid of the backup route maybe.

solutionDrive-Alt: @stefandoorn for me this would be perfectly fine, as long as the information how to do this is relatively easy to find.

stefandoorn:
Made some changes. I think this is good to go for me to test in production. We need Behat tests btw to check whether all the exports are actually returning what we want, but I'm not able to write them on short notice. I'm able to test on short notice in production though for now.

src/Listener/ExportButtonGridListener.php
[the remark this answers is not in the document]
lsmith77: can we add a comment into the code here that we should look into some approach to validate the structure of this parameter?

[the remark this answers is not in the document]
stefandoorn: Done!

README.md
```diff
@@ -320,7 +320,7 @@ Optional add the routing:
 
 ```yaml
 app_export_data_country:
-    path: /export/sylius.country/{format}
+    path: /admin/export/sylius.country/{format}
```
lsmith77: good idea .. but shouldn’t we do this more consistently then?

stefandoorn: Sure. Shall we tackle that in a separate issue?

lsmith77: https://github.com/FriendsOfSylius/SyliusImportExportPlugin/pull/81/files#diff-04c6e90faac2675aa89e2176d2eec7d8R263

lsmith77: https://github.com/FriendsOfSylius/SyliusImportExportPlugin/pull/81/files#diff-a35a2fe98a35bdaf457f33ada44dc860R9

PR #82, opened 2018-03-18
[Export] Add feature tests for export UI on countries (first merge #78)

stefandoorn:
First features to check links in admin grid.

### DIFF
--- a/tests/Behat/Context/CountriesContext.php
+++ b/tests/Behat/Context/CountriesContext.php
@@ -5,11 +5,14 @@ declare(strict_types=1);
 namespace Tests\FriendsOfSylius\SyliusImportExportPlugin\Behat\Context;
 
 use Behat\Behat\Context\Context;
+use Behat\Mink\Session;
 use PHPUnit\Framework\Assert;
 use Sylius\Behat\Context\Transform\CountryContext;
+use Sylius\Behat\Page\SymfonyPage;
+use Symfony\Component\Routing\RouterInterface;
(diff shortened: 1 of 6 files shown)


tests/Behat/Context/CountriesContext.php
```diff
@@ -57,4 +104,15 @@ public function theCountryShouldAppearInTheRegistry($countryname)
             sprintf('Country with name "%s" does not exist', $countryname)
         );
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function getDefinedElements()
```
solutionDrive-Alt: does this have to be in the context? (I think that's the reason for extending SymfonyPage?) Never saw a context extending SymfonyPage before and ask myself, if this method rather belongs to the IndexPage-Object, or is there a specific reason that it is in the context?

stefandoorn: It's used by the `getElement` calls done higher in the file.. I 'stole' this from Sylius, so I'm not sure. Might be better to place them in the IndexPage, if that works. Is it ok to file this as a new issue for improvement?

solutionDrive-Alt: yeah, just was surprised to see the context extending SymfonyPage. For now it is good for me, since I don't really know what's the better take on this at the moment.

stefandoorn: Maybe @pamil can hint us to a better solution :-)

PR #84, opened 2018-03-25
Exporter field transformers before writing data to file

stefandoorn:
Currently only 'DateTime' -> string as default & one that adjusts Numeric values to money format for specific fields (currently only the 'Total' field as used in Orders).


stefandoorn:
This one can be reviewed @solutionDrive-Alt.

stefandoorn:
Danke schön :-)

PR #98, opened 2018-04-27
Add option to skip checking for the zip extension

stefandoorn:
Even though I introduced the check myself, I also had some issues installing the zip extension locally at the moment. This extra option allows to skip the check, having the services available so DI won't throw an exception on each request.

### DIFF
--- a/src/DependencyInjection/FOSSyliusImportExportExtension.php
+++ b/src/DependencyInjection/FOSSyliusImportExportExtension.php
@@ -50,7 +50,8 @@ class FOSSyliusImportExportExtension extends Extension
             $loader->load('services_import_excel.yml');
         }
 
-        if (class_exists(self::CLASS_CSV_WRITER) && extension_loaded('zip')) {
+        if (class_exists(self::CLASS_CSV_WRITER) &&
+          (true === $config['exporter']['zip_extension_optional'] || extension_loaded('zip'))) {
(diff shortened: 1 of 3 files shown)


src/DependencyInjection/FOSSyliusImportExportExtension.php
```diff
@@ -50,7 +50,8 @@ public function load(array $configs, ContainerBuilder $container)
             $loader->load('services_import_excel.yml');
         }
 
-        if (class_exists(self::CLASS_CSV_WRITER) && extension_loaded('zip')) {
+        if (class_exists(self::CLASS_CSV_WRITER) &&
```
solutionDrive-Alt: I just saw that we use the CSV_WRITER - Constant here.

src/DependencyInjection/FOSSyliusImportExportExtension.php
```diff
@@ -50,7 +50,8 @@ public function load(array $configs, ContainerBuilder $container)
...
+          (true === $config['exporter']['zip_extension_optional'] || extension_loaded('zip'))) {
```
solutionDrive-Alt: Do we have to load the services at all, if we don't have the zip extension?

stefandoorn:
I think I can close this one and use another method, will check it soon.

PR #102, opened 2018-05-18
CLI export fix

enesaktay:
- Added concatenation of 'sylius.' with the exporter as the exporter got renamed and needs a sylius infront now.

- Since the exporter has sylius infront now we have to ignore the first field of $exporter[]

### DIFF
--- a/src/Command/ExportDataCommand.php
+++ b/src/Command/ExportDataCommand.php
@@ -65,7 +65,7 @@ final class ExportDataCommand extends Command
             $this->listExporters($input, $output, $message);
         }
         $format = $input->getOption('format');
-        $name = ExporterRegistry::buildServiceName(('sylius.' . $exporter), $format);
+        $name = ExporterRegistry::buildServiceName('sylius.' . $exporter, $format);
 
         if (!$this->exporterRegistry->has($name)) {
             $message = sprintf(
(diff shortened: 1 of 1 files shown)


src/Command/ExportDataCommand.php
```diff
@@ -65,7 +65,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
             $this->listExporters($input, $output, $message);
         }
         $format = $input->getOption('format');
-        $name = ExporterRegistry::buildServiceName($exporter, $format);
+        $name = ExporterRegistry::buildServiceName(('sylius.' . $exporter), $format);
```
stefandoorn: I don't think you need the additional `()` around it :-)

src/Command/ExportDataCommand.php
```diff
@@ -118,7 +118,7 @@ private function listExporters(
         $exporters = [];
         foreach ($all as $exporter) {
             $exporter = explode('.', $exporter);
-            $exporters[$exporter[0]][] = $exporter[1];
+            $exporters[$exporter[1]][] = $exporter[2];
```
stefandoorn: I know it now also doesn't have a comment, but a comment in here would probably clarify a lot what's happening in here :-) E.g. an example of how an example string consist and is being passed on afterwards.

PR #106, opened 2018-05-24
fixed ids of export on ui

enesaktay:
old
is now:
new


src/Controller/ExportDataController.php
```diff
@@ -89,8 +89,16 @@ private function exportData(Request $request, string $exporter, string $format,
         /** @var ResourceExporterInterface $service */
         $service = $this->registry->get($name);
 
-        $resources = $this->findResources($configuration, $this->findRepository($resource));
```
lsmith77: the method should then be removed

solutionDrive-Alt: this will most likely be also true for the findRepository() Method, if this line is the only reference to it.

stefandoorn:
There’s no tests for grid filter UI exports yet, so therefore tests pass also when removing the functionality.

enesaktay:
Oh wow, you are right.. my bad!
Thanks

enesaktay:
Redid the fix so its open for review again. Now it supports for grid filtering again :)

stefandoorn:
Am I correct to understand that the second page was loaded twice instead of the first + second one? 

Can we add tests to cover this issue + the fix that's provided? So it won't happen again.

lsmith77:
yeah good idea .. should be a good exercise before tackling bigger topics too :)

PR #103, opened 2018-05-18
Added an export of orders

enesaktay:

PR #107, opened 2018-05-24
added the export for orders

enesaktay:
#103 had unwanted commits in it so this is the same PR without the unwanted commits


src/Resources/config/services_export_csv.yml
```diff
@@ -17,3 +17,14 @@ services:
...
+    sylius.exporter.orders.csv:
+        class: FriendsOfSylius\SyliusImportExportPlugin\Exporter\ResourceExporter
+        arguments:
+            - "@sylius.exporter.csv_writer"
+            - "@sylius.exporter.pluginpool.orders"
+            - ["Number", "State", "Product_list" , "Total" ,  "Currency_code", "Checkout_completed_at", "Checkout_state", "Payment_state", "Shipping_state", "Token_value", "Gender", "Full_name", "Email", "Telephone", "Shipping_address", "Billing_address" , "Customer_ip", "Notes"]
+            - "@sylius.exporters_transformer_pool"
+        tags:
+            - { name: sylius.exporter, type: sylius.order, format: csv }
+
```
solutionDrive-Alt: Here is a blank line missing at the end of the file

src/Resources/config/services_export_excel.yml
```diff
@@ -17,3 +17,14 @@ services:
...
+    sylius.exporter.orders.xlsx:
+        class: FriendsOfSylius\SyliusImportExportPlugin\Exporter\ResourceExporter
+        arguments:
+            - "@sylius.exporter.excel_writer"
+            - "@sylius.exporter.pluginpool.orders"
+            - ["Number", "State", "Product_list" , "Total" ,  "Currency_code", "Checkout_completed_at", "Checkout_state", "Payment_state", "Shipping_state", "Token_value", "Gender", "Full_name", "Email", "Telephone", "Shipping_address", "Billing_address" , "Customer_ip", "Notes"]
+            - "@sylius.exporters_transformer_pool"
+        tags:
+            - { name: sylius.exporter, type: sylius.order, format: xlsx }
+
```
solutionDrive-Alt: same here.

src/Exporter/Plugin/OrderResourcePlugin.php
```diff
@@ -0,0 +1,102 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function init(array $idsToExport): void
+    {
+        parent::init($idsToExport);
+
+        /** @var OrderInterface $resource */
+        foreach ($this->resources as $resource) {
```
solutionDrive-Alt: could you refactor this foreach a bit, so it gets a bit smaller? I would recommend to move the contents of the different if-conditionals to private methods, that would make the code of this init-method a bit clearer to understand.

src/Exporter/Plugin/OrderResourcePlugin.php
```diff
@@ -0,0 +1,102 @@
...
+                    $billingAddress->getCity() ? $billingAddress->getCity() : '',
+                    $billingAddress->getPostcode() ? $billingAddress->getPostcode() : '',
+                    $billingAddress->getCountryCode() ? $billingAddress->getCountryCode() : ''
+                ));
+            }
+
+            $items = [];
+
+            /** @var OrderItemInterface $orderItem */
+            foreach ($resource->getItems() as $orderItem) {
```
solutionDrive-Alt: Also the following both foreach-loops could be relatively easy extracted to a private-method like 'addOrderItemInformation' or something like that, I think.

src/Exporter/Plugin/OrderResourcePlugin.php
```diff
@@ -0,0 +1,102 @@
...
+use Symfony\Component\PropertyAccess\PropertyAccessorInterface;
+
+class OrderResourcePlugin extends ResourcePlugin
+{
+    /**
+     * @param RepositoryInterface $repository
+     * @param PropertyAccessorInterface $propertyAccessor
+     * @param EntityManagerInterface $entityManager
+     */
+    public function __construct(
```
lsmith77: this looks redundant and covered by the parent constructor

src/Resources/config/services_export_csv.yml
enesaktay: For me at row 30 I have a blank line?

solutionDrive-Alt: Hmm, I don't know what's then the problem there. Github does mention it.
bildschirmfoto 2018-05-25 um 10 48 45
Maybe there is some Whitespace in this line?

enesaktay:
Thanks for the reviews and help, I did the requested changes.. its open for review again :)

src/Exporter/Plugin/OrderResourcePlugin.php
```diff
@@ -0,0 +1,130 @@
...
+            $items = $this->getItemsAndCount($resource);
+
+            $this->addOrderItemData($items, $resource);
+        }
+    }
+
+    /**
+     * @param OrderInterface $resource
+     */
+    private function addCustomerData(OrderInterface $resource)
```
solutionDrive-Alt: bit nitpicking, but can you add a TypeHint for the retrun-value of all added private Methods?
e.g. `private function addCustomerData(OrderInterface $resource): void`

We should typeHint as much as we can... :-)

src/Exporter/Plugin/OrderResourcePlugin.php
```diff
@@ -0,0 +1,130 @@
...
+            /** @var ProductVariantInterface $variant */
+            $variant = $orderItem->getVariant();
+            /** @var ProductInterface $product */
+            $product = $variant->getProduct();
+
+            if (!isset($items[$product->getName()])) {
+                $items[$product->getName()] = 0;
+            }
+
+            $items[$product->getName()] += $orderItem->getQuantity();
```
stefandoorn: How does this with products with the same name but not the same product ID? Basically different product, but the count on the name gets upped.

src/Exporter/Plugin/OrderResourcePlugin.php
```diff
@@ -0,0 +1,130 @@
...
+    /**
+     * @param OrderInterface $resource
+     */
+    private function addShippingAddressData(OrderInterface $resource): void
+    {
+        $shippingAddress = $resource->getShippingAddress();
+        if (null !== $shippingAddress) {
+            $this->addDataForResource($resource, 'Shipping_address', sprintf(
+                '%s, %s, %s, %s, %s',
+                $shippingAddress->getFullName() ? $shippingAddress->getFullName() : '',
```
stefandoorn: Can't we use `??` here?

src/Exporter/Plugin/OrderResourcePlugin.php
```diff
@@ -0,0 +1,130 @@
...
+        }
+    }
+
+    /**
+     * @param OrderInterface $resource
+     */
+    private function addShippingAddressData(OrderInterface $resource): void
+    {
+        $shippingAddress = $resource->getShippingAddress();
+        if (null !== $shippingAddress) {
```
stefandoorn: What about early returns? `if (null !== $shippingAddress) { return; }`

src/Exporter/Plugin/OrderResourcePlugin.php
lsmith77: yeah .. this is risky .. should use something unique for the key

src/Exporter/Plugin/OrderResourcePlugin.php
lsmith77: +1 for return early

src/Exporter/Plugin/OrderResourcePlugin.php
solutionDrive-Alt: Since I already merged this, I think we should open an issue for that. I totally overseen this..

src/Exporter/Plugin/OrderResourcePlugin.php
lsmith77: I am not sure if anything is even needed here .. I assume the return value is either a string or null and then it will just silently get cast to an empty string.

stefandoorn: @enesaktay is this something we can make easier to read/use?

enesaktay: @stefandoorn could potentially remove the ternary operators since with row 58 we are already checking if shippingadress exists
other than that i could also put sprintf inside another method since it is used 2 times in this class

stefandoorn: Indeed, might be worth trying if we need the operator at all as the methods probably return useful information anyway..

PR #109, opened 2018-05-29
JSON import/export functionality

enesaktay:
export is done now, can be reviewed
import is tbd


src/DependencyInjection/FOSSyliusImportExportExtension.php
```diff
@@ -21,6 +22,9 @@ class FOSSyliusImportExportExtension extends Extension
     private const CLASS_EXCEL_READER = ExcelReaderFactory::class;
     private const CLASS_EXCEL_WRITER = ExcelWriter::class;
 
+//    private const CLASS_JSON_READER = Reader::class;
```
stefandoorn: Oeps :)

src/Exporter/ResourceExporter.php
```diff
@@ -16,17 +16,17 @@ class ResourceExporter implements ResourceExporterInterface
     /**
      * @var array
      */
-    private $resourceKeys;
+    protected $resourceKeys;
```
stefandoorn: Why do they need to become protected? Isn't filtering done in a later stage actually?

src/Exporter/JsonResourceExporter.php
```diff
@@ -0,0 +1,44 @@
...
+     */
+    public function export(array $idsToExport, string $filename = 'exportfile'): void
+    {
+        $this->pluginPool->initPlugins($idsToExport);
+
+        foreach ($idsToExport as $id) {
+            $this->writeDataForId((string) $id);
+        }
+
+        $myfile = fopen($filename, "w") or die("Unable to open file!");
```
lsmith77: we should handle the error condition with an exception.

btw the same should be done here I guess https://github.com/FriendsOfSylius/SyliusImportExportPlugin/blob/master/src/Writer/CsvWriter.php#L37

enesaktay:
@lsmith77 i was thinking the same, but since #110 is open i just hotfixed it like that 👍

enesaktay:
removed WIP as this branch is done now and open for review

composer.json
```diff
@@ -35,7 +37,9 @@
         "phpunit/phpunit": "^5.6",
         "se/selenium-server-standalone": "^2.52",
         "sylius-labs/coding-standard": "^1.0",
-        "symplify/easy-coding-standard": "^2.4"
+        "symplify/easy-coding-standard": "^2.4",
+        "dcarbone/json-writer-plus": "^0.3.0",
```
stefandoorn: Are there more stable libraries available to support JSON export & import? It might be good stuff, but no stable release yet apparently.

features/export/cli/exporting_countries_to_json_via_cli.feature
```diff
@@ -0,0 +1,13 @@
+@managing_countries
+Feature: exporting countries to json-file
+  In order to have my countries exported to an external target
+  As a Developer
+  I want to be able to export country data to json file from the commandline
+
+  Background:
+    Given I have a working command-line-interface
```
stefandoorn: `command-line interface` :-)

features/export/cli/exporting_countries_to_json_via_cli.feature
```diff
@@ -0,0 +1,13 @@
+@managing_countries
+Feature: exporting countries to json-file
```
stefandoorn: `JSON file`?

features/export/cli/exporting_countries_to_json_via_cli.feature
```diff
@@ -0,0 +1,13 @@
...
+  As a Developer
+  I want to be able to export country data to json file from the commandline
+
+  Background:
+    Given I have a working command-line-interface
+
+  @cli_importer_exporter
+  Scenario: Exporting countries to json-file
+    When I export "country" data as "json" to the file "countries_export.json" with the cli-command
+    Then I should see "Exported" in the output
```
stefandoorn: Would be great to have an extra check that the file actually exists also on the file system.

src/Exporter/JsonResourceExporter.php
```diff
@@ -0,0 +1,48 @@
...
+     */
+    public function export(array $idsToExport, string $filename = 'exportfile'): void
+    {
+        $this->pluginPool->initPlugins($idsToExport);
+
+        foreach ($idsToExport as $id) {
+            $this->writeDataForId((string) $id);
+        }
+
+        $myfile = fopen($filename, 'w');
```
lsmith77: just a minor thing .. I would prefer `$file` over `$myfile` .. but I can also accept `$myfile` if you prefer this naming

features/export/cli/exporting_countries_to_json_via_cli.feature
```diff
@@ -0,0 +1,13 @@
+@managing_countries
+Feature: exporting countries to json-file
+  In order to have my countries exported to an external target
+  As a Developer
```
stefandoorn: `As a developer`

features/import/cli/importing_countries_from_json_via_cli.feature
```diff
@@ -0,0 +1,16 @@
+@managing_countries
+Feature: Importing countries from json with the command-line-interface
+    In order to have my countries from external source
+    As an Developer
```
stefandoorn: `As a developer`

features/import/ui/importing_countries_from_json.feature
```diff
@@ -0,0 +1,17 @@
...
+    Given I am logged in as an administrator
+
+  @ui
+  Scenario: Importing countries based on a valid json-file
+    When I open the country admin index page
+    And I import country data from "countries.json" json file
+    Then I should see a notification that the import was successful
+    And I should see 2 countries in the list
+    And the country "Germany" should appear in the registry
+    And the country "Switzerland" should appear in the registry
```
stefandoorn: Nice! :)

src/Exporter/JsonResourceExporter.php
stefandoorn: It seems we don't really stream to the file, but it's more like doing a single write statement. Could be replaced with `file_put_contents` therefore I think.

src/Exporter/ResourceExporter.php
lsmith77: `JsonResourceExporter extends ResourceExporter` is the reason (also the the other attributes).
We might need to think about adding accessor methods instead of maybe refactoring a bit more as atm I guess too much is still build around portphp .. but I am not sure if this should happen within this PR

src/Importer/JsonResourceImporter.php
```diff
@@ -0,0 +1,68 @@
...
+
+declare(strict_types=1);
+
+namespace FriendsOfSylius\SyliusImportExportPlugin\Importer;
+
+use Doctrine\Common\Persistence\ObjectManager;
+use FriendsOfSylius\SyliusImportExportPlugin\Processor\ResourceProcessorInterface;
+use pcrov\JsonReader\JsonReader;
+
+class JsonResourceImporter extends ResourceImporter
```
stefandoorn: Can this one be `final`?

src/Importer/JsonResourceImporter.php
```diff
@@ -0,0 +1,68 @@
...
+        $this->failOnIncomplete = $failOnIncomplete;
+        $this->stopOnFailure = $stopOnFailure;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function import(string $fileName): ImporterResultInterface
+    {
+        $reader = $this->reader;
```
stefandoorn: No need for local variable here :)

src/Importer/JsonResourceImporter.php
```diff
@@ -0,0 +1,68 @@
...
+        foreach ($reader->value() as $i => $row) {
+            $breakBool = $this->importData($i, $row);
+            if ($breakBool) {
+                break;
+            }
+        }
+
+        $reader->close(); // Close the reader
+
+        if ($batchCount) {
```
stefandoorn: Will this ever be called at all? (not sure about the use also)

src/Importer/JsonResourceImporter.php
```diff
@@ -0,0 +1,68 @@
...
+        $this->result->start();
+
+        $batchCount = 0;
+
+        $depth = $reader->depth(); // Check in a moment to break when the array is done
+
+        $reader->read(); // Step to the first element
+
+        foreach ($reader->value() as $i => $row) {
+            $breakBool = $this->importData($i, $row);
```
stefandoorn:
```
if (!$this->importData($i, $row) {
    break;
}
```

src/Importer/JsonResourceImporter.php
```diff
@@ -0,0 +1,68 @@
...
+    {
+        $reader = $this->reader;
+
+        $reader->open($fileName);
+
+        $this->result->start();
+
+        $batchCount = 0;
+
+        $depth = $reader->depth(); // Check in a moment to break when the array is done
```
stefandoorn: Is `$depth` used at all?

src/Importer/ResourceImporter.php
```diff
@@ -62,40 +65,58 @@ public function import(string $fileName): ImporterResultInterface
...
+            ++$this->batchCount;
+            if ($this->batchSize && $this->batchCount === $this->batchSize) {
+                $this->objectManager->flush();
+                $this->batchCount = 0;
+            }
+        } catch (ItemIncompleteException $e) {
+            if ($this->failOnIncomplete) {
+                $this->result->failed($i);
+                if ($this->stopOnFailure) {
+                    $breakBool = true;
```
stefandoorn: Can `return true` in here directly?

src/Importer/ResourceImporter.php
```diff
@@ -62,40 +65,58 @@ public function import(string $fileName): ImporterResultInterface
...
+                }
+            } else {
+                $this->result->skipped($i);
+            }
+        } catch (ImporterException $e) {
+            $this->result->failed($i);
+            if ($this->stopOnFailure) {
+                $breakBool = true;
```
stefandoorn: Can `return true` in here directly?

src/Importer/ResourceImporter.php
```diff
@@ -62,40 +65,58 @@ public function import(string $fileName): ImporterResultInterface
...
-                    }
-                } else {
-                    $this->result->skipped($i);
-                }
-            } catch (ImporterException $e) {
-                $this->result->failed($i);
-                if ($this->stopOnFailure) {
-                    break;
-                }
+            $breakBool = $this->importData($i, $row);
```
stefandoorn: I would see if there's a way to get around these `$breakBool` variables, but break/return earlier :)

src/Importer/JsonResourceImporter.php
```diff
@@ -0,0 +1,68 @@
...
+use FriendsOfSylius\SyliusImportExportPlugin\Processor\ResourceProcessorInterface;
+use pcrov\JsonReader\JsonReader;
+
+class JsonResourceImporter extends ResourceImporter
+{
+    /** @var JsonReader */
+    private $reader;
+
+    public function __construct(
+        JsonReader $reader,
```
stefandoorn: Is there an interface available for this one?

composer.json
lsmith77: my concern is that this library is not MIT licensed which means it would add another license to the mix.

but I am not sure if we even need it .. basically we could just collect all the data into one big array and then use `file_put_contents(json_encode($data))`

src/Writer/CsvWriter.php
```diff
@@ -34,7 +34,12 @@ public function write(array $data): void
      */
     public function setFile(string $filename): void
     {
-        $this->writer->setStream(fopen($filename, 'w+'));
+        $myfile = fopen($filename, 'w+');
```
stefandoorn: I would go with `$file` (cc @lsmith77). It's yours anyway ;-)

src/Writer/JsonWriter.php
```diff
@@ -0,0 +1,57 @@
+<?php
+
+declare(strict_types=1);
+
+namespace FriendsOfSylius\SyliusImportExportPlugin\Writer;
+
+use DCarbone\JsonWriterPlus;
+
+class JsonWriter implements WriterInterface
```
stefandoorn: Can this be `final`?

src/Writer/JsonWriter.php
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * @param JsonWriterPlus $writer
+     */
+    public function __construct(
+        JsonWriterPlus $writer
+    ) {
+        $this->writer = $writer;
+        $this->writer->startJson();
+        $this->writer->writeStartArray();
```
stefandoorn: Don't think it's good to do this in the constructor. As soon as Symfony creates the object (for DI e.g.), it's being run.

composer.json
lsmith77: I mean a custom write would only be interesting if it could "stream" which doesn't seem to be the case (unless I overlooked something).

features/export/cli/exporting_countries_to_json_via_cli.feature
lsmith77: I guess we should add something like this to the other formats too.

src/Importer/JsonResourceImporter.php
lsmith77: I think this relates to also `$batchCount` .. it seems like there was an intention to write to the DB in batches but this was not implemented?

src/Writer/JsonWriter.php
lsmith77: if we drop `JsonWriterPlus` then this issue will solve itself :)

stefandoorn: Haha yes. I think it's still good to keep the JsonWriter as a class with an internal dependency+interface, so we can always switch to an external library for whatever reason, though (other suggestion about `json_encode` is good I think).

enesaktay:
removed both jsonreader and jsonwriter libraries and made the asked changes

src/Importer/ResourceImporter.php
```diff
@@ -62,40 +65,59 @@ public function import(string $fileName): ImporterResultInterface
...
+
+    /**
+     * @param int $i
+     * @param array $row
+     *
+     * @return bool
+     */
+    public function importData($i, $row): bool
+    {
+//        $breakBool = false;
```
lsmith77: this can now be removed

lsmith77:
looks great .. but not sure why the tests are failing ..

stefandoorn:
Seems related to rename of `command-line interface` which not seems to be renamed everywhere.

enesaktay:
@stefandoorn thanks, now the tests run through :)
**cc** @lsmith77

src/Writer/CsvWriter.php
```diff
@@ -34,7 +34,12 @@ public function write(array $data): void
      */
     public function setFile(string $filename): void
     {
-        $this->writer->setStream(fopen($filename, 'w+'));
+        $file = fopen($filename, 'w+');
```
lsmith77: this is actually a bug fix unrelated to this PR. it would have been better to keep this separate .. but oh well :)

phpstan.neon
```diff
@@ -10,3 +10,5 @@ parameters:
     ignoreErrors:
         - '/Parameter #1 $configuration of method Symfony\Component\DependencyInjection\Extension\Extension::processConfiguration() expects Symfony\Component\Config\Definition\ConfigurationInterface, Symfony\Component\Config\Definition\ConfigurationInterface|null given./'
         - '/Parameter #1 $iterator of function iterator_to_array expects Traversable, array|Traversable given./'
+        - '#[a-zA-Z0-9\\_]JsonResourceExporter::__construct()#'
```
lsmith77: created https://github.com/FriendsOfSylius/SyliusImportExportPlugin/issues/119

PR #110, opened 2018-05-29
Fix CLI Excel export creating files with zero bytes

### LINKED ISSUE #104: exporting xlsx via cli creates a file with zero bytes

enesaktay:
fixes #104


src/Writer/WriterInterface.php
```diff
@@ -20,4 +20,9 @@ public function setFile(string $filename): void;
      * @return string
      */
     public function getFileContent(): string;
+
+    /**
+     * Wrap up the writer after all items have been written
```
stefandoorn: Maybe more clear: `Clear the writer after all items have been written`?

src/Writer/ExcelWriter.php
```diff
@@ -86,4 +86,12 @@ private function prepare(): void
             $this->writer->prepare();
         }
     }
+
+    /**
+     * Wrap up the writer after all items have been written
```
stefandoorn: Can use `@inheritdoc` here

src/Exporter/ResourceExporterInterface.php
```diff
@@ -20,4 +20,9 @@ public function setExportFile(string $filename): void;
      * @return string
      */
     public function getExportedData(): string;
+
+    /**
+     * Wrap up the writer after all items have been written
+     */
+    public function finish(): void;
```
stefandoorn: The downside of adding extra methods to the interface is that any implementation needs to provide them. The less, the better sometimes. WDYT @lsmith77 @solutionDrive-Alt on this PR?

src/Writer/WriterInterface.php
enesaktay: The description has been copied from the portphp writer interface, but if you think clear fits more I can use clear :)

src/Exporter/ResourceExporterInterface.php
lsmith77: well for some it may be an empty implementation .. but I see no way around this, at least given how portphp is currently setup.

solutionDrive-Alt: I'm a bit out of context here. Is the reason why we need this method somewhere documented? I heard about the problem, that zero-byte files are generated, but I'm not entirely sure what it has to do with this finish-method...

enesaktay: @solutionDrive-Alt the excel cli export didnt call the finish method of the phpexcel writer, this caused the file to be created with zero-bytes

stefandoorn: Maybe we should look into some refactoring where we share more between CLI and UI exports, in order to make them as least different as possible. Maybe I'm missing something, but I think it would be a good improvement.

spec/Writer/CsvWriterSpec.php
```diff
@@ -36,4 +36,12 @@ function it_delegates_the_data_to_the_wrapped_writer(Writer $csvWriter)
         $csvWriter->writeItem($data)->shouldBeCalled();
         $this->write($data);
     }
+
+    function it_finishes_the_file_creation_when_we_get_the_contents(Writer $csvWriter)
+    {
+        $csvWriter->setCloseStreamOnFinish(true)->shouldBeCalled();
```
lsmith77: it seems like phpspec needs this .. but I would hope that there is some way to not have to mock this method since its an implementation detail we are not interested in testing here.

src/Exporter/ResourceExporterInterface.php
lsmith77: this relates a bit to https://github.com/FriendsOfSylius/SyliusImportExportPlugin/issues/119 .. right now we are too heavily bound to portphp

Issue #112, opened 2018-05-29
Use more early returns in order export


stefandoorn:
https://github.com/FriendsOfSylius/SyliusImportExportPlugin/pull/107#pullrequestreview-123758419

cc @enesaktay

stefandoorn:
@enesaktay What I meant is something like this:

```php
private function addBillingAddressData(OrderInterface $resource): void
    {
        $billingAddress = $resource->getBillingAddress();
        if (null === $billingAddress) {
            return;
        }

            $this->addDataForResource($resource, 'Billing_address', sprintf(
                '%s, %s, %s, %s, %s',
                $billingAddress->getFullName() ? $billingAddress->getFullName() : '',
                $billingAddress->getStreet() ? $billingAddress->getStreet() : '',
                $billingAddress->getCity() ? $billingAddress->getCity() : '',
                $billingAddress->getPostcode() ? $billingAddress->getPostcode() : '',
                $billingAddress->getCountryCode() ? $billingAddress->getCountryCode() : ''
            ));
    }
```

Instead of:

```php
private function addBillingAddressData(OrderInterface $resource): void
    {
        $billingAddress = $resource->getBillingAddress();
        if (null !== $billingAddress) {
            $this->addDataForResource($resource, 'Billing_address', sprintf(
                '%s, %s, %s, %s, %s',
                $billingAddress->getFullName() ? $billingAddress->getFullName() : '',
                $billingAddress->getStreet() ? $billingAddress->getStreet() : '',
                $billingAddress->getCity() ? $billingAddress->getCity() : '',
                $billingAddress->getPostcode() ? $billingAddress->getPostcode() : '',
                $billingAddress->getCountryCode() ? $billingAddress->getCountryCode() : ''
            ));
        }
    }
```

enesaktay:
done #113

PR #113, opened 2018-05-30
adjustments made to the order export

### LINKED ISSUE #111: Improve fetching count in order exporter
https://github.com/FriendsOfSylius/SyliusImportExportPlugin/pull/107#pullrequestreview-123758038

cc @enesaktay

enesaktay:
fixes #111 and https://github.com/FriendsOfSylius/SyliusImportExportPlugin/issues/112


src/Exporter/Plugin/OrderResourcePlugin.php
```diff
@@ -101,11 +101,13 @@ private function getItemsAndCount(OrderInterface $resource): array
             /** @var ProductInterface $product */
             $product = $variant->getProduct();
 
-            if (!isset($items[$product->getName()])) {
-                $items[$product->getName()] = 0;
+            if (!isset($items[$product->getId()])) {
+                $items[$product->getId()] = [
+                    'itemName' => $product->getName(),
```
stefandoorn: Nitpicking, but I think you can use `name` and `count`, as it's already clear you are having items :-)

enesaktay: @stefandoorn cool! fixed it
@solutionDrive-Alt I understand your statement, I didn't fully understand if you wanted me to change it or if its ok like this?

stefandoorn:
Looks good :-)

enesaktay:
@stefandoorn what do you think of the changes in the last push as we were talking about it in #107 ?

lsmith77:
@enesaktay I am not sure if I would extract the values from the objects before handing it off .. wouldn't it make more sense to simply pass the entire object (graph) to this service and then letting the service determine the relevant properties to include?

enesaktay:
@lsmith77 donezo, hope I did what you were thinking of

lsmith77:
yeah ..

not sure though about the name.

instead of `AddressInfo` I would just use `Address`

I would probaly call the class `DefaultAddressConcatenation` and the interface `AddressConcatenationInterface` ..

src/Service/DefaultAddressConcatenation.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
+
+namespace FriendsOfSylius\SyliusImportExportPlugin\Service;
+
+use Sylius\Component\Core\Model\AddressInterface;
+
+class DefaultAddressConcatenation implements AddressConcatenationInterface
```
stefandoorn: Final class?

src/Service/DefaultAddressConcatenation.php
```diff
@@ -0,0 +1,29 @@
...
+{
+    /**
+     * @param AddressInterface $address
```
stefandoorn: I think it's ok here to just return directly :)

src/Service/DefaultAddressConcatenation.php
```diff
@@ -0,0 +1,29 @@
...
+     *
+     * @return string
```
stefandoorn: Can replace the whole comment with `@inheritdoc` here, as the interface has the same comment :)

src/Service/AddressConcatenationInterface.php
```diff
@@ -0,0 +1,17 @@
...
+use Sylius\Component\Core\Model\AddressInterface;
+
+interface AddressConcatenationInterface
+{
+    /**
+     * @param AddressInterface $address
+     *
+     * @return string
+     */
+    public function getConcatenatedAddressString(AddressInterface $address): string;
```
stefandoorn: `getConcatenatedAddress` instead of `getConcatenatedAddressString`?

stefandoorn:
I'm not sure we want to `composer.lock` file in this PR, cc @solutionDrive-Alt @lsmith77

src/Service/AddressConcatenationInterface.php
lsmith77: I guess since we now have a dedicated class .. maybe just getString() .. or __toString()?

but then the classes and interfaces should maybe we renamed again ..
something like AddressStringifier and StringifierInterface?

lsmith77:
@stefandoorn it appears we need the composer.lock to make scrutinizer happy .. but we can move the composer.lock into a separate PR

src/Exporter/Plugin/OrderResourcePlugin.php
```diff
@@ -55,16 +60,17 @@ private function addCustomerData(OrderInterface $resource): void
...
-                $shippingAddress->getPostcode() ? $shippingAddress->getPostcode() : '',
-                $shippingAddress->getCountryCode() ? $shippingAddress->getCountryCode() : ''
-            ));
+
+        if (null === $shippingAddress) {
+            return;
         }
+
+        /** @var AddressConcatenationInterface $addressConcatenation */
+        $addressConcatenation = new DefaultAddressConcatenation();
```
stefandoorn: Can we inject this one as dependency via the constructor?

stefandoorn:
Small comment. And, will you be able to add tests for the stuff happening in this PR? Quite testable I think + important features.

src/Exporter/Plugin/OrderResourcePlugin.php
```diff
@@ -4,6 +4,8 @@
 
 namespace FriendsOfSylius\SyliusImportExportPlugin\Exporter\Plugin;
 
+use FriendsOfSylius\SyliusImportExportPlugin\Service\AddressConcatenationInterface;
+use FriendsOfSylius\SyliusImportExportPlugin\Service\DefaultAddressConcatenation;
```
stefandoorn: Is this `use` statement actually used?

src/Exporter/Plugin/OrderResourcePlugin.php
stefandoorn: (and just typehint the interface)

src/Service/AddressConcatenationInterface.php
```diff
@@ -0,0 +1,17 @@
...
+
+declare(strict_types=1);
+
+namespace FriendsOfSylius\SyliusImportExportPlugin\Service;
+
+use Sylius\Component\Core\Model\AddressInterface;
+
+interface AddressConcatenationInterface
+{
+    /**
```
stefandoorn: @lsmith77 @solutionDrive-Alt What will we decide about this kind of docblocks? The info in it doesn't add anything looking at the method definition. I'm indifferent about it, but it seems more and more accepted to leave them out (e.g. latest versions of PHPStan/CS are promoting this I think).

lsmith77: I personally would also leave out phpdoc for things already expressed in code .. however I think in the past we have added phpdoc either way .. so imho we should do a separate PR with a cleanup.

spec/Exporter/Plugin/OrderResourcePluginSpec.php
```diff
@@ -0,0 +1,110 @@
...
+    function it_should_add_customer_data(
+        RepositoryInterface $repository,
+        EntityManagerInterface $entityManager,
+        OrderInterface $resource,
+        PropertyAccessorInterface $propertyAccessor,
+        ClassMetaData $classMetadata
+    ) {
+        $idsToExport = [1];
+
+        $repository->findBy(
```
lsmith77: imho this is a but too much whitespace .. think it would be sufficiently readable if the parameters would be on the same line.

ie.

```
$repository->findBy(['id' => $idsToExport])
    ->willReturn([$resource]);
```

obviously if we ever refactor this and need more parameters it would create additional noise in the diff. but until then I would prefer it be more compact to reduce need for scrolling.

spec/Exporter/Plugin/OrderResourcePluginSpec.php
```diff
@@ -0,0 +1,110 @@
...
+//        $resource->getLocaleCode()->shouldBeCalled();
+//        $resource->getCheckoutState()->shouldBeCalled();
+//        $resource->getPaymentState()->shouldBeCalled();
+//        $resource->getShippingState()->shouldBeCalled();
+//        $resource->getTokenValue()->shouldBeCalled();
+//        $resource->getCustomerIp()->shouldBeCalled();
+//
+//        $propertyAccessor->isReadable($resource, 'Number')->willReturn('000000001');
+
+//        $this->init($idsToExport);
```
lsmith77: I guess this part is WIP

src/Service/AddressConcatenationInterface.php
lsmith77: created https://github.com/FriendsOfSylius/SyliusImportExportPlugin/issues/118

enesaktay:
@lsmith77 can be merged :)

phpstan.neon
```diff
@@ -10,5 +10,6 @@ parameters:
     ignoreErrors:
         - '/Parameter #1 $configuration of method Symfony\Component\DependencyInjection\Extension\Extension::processConfiguration() expects Symfony\Component\Config\Definition\ConfigurationInterface, Symfony\Component\Config\Definition\ConfigurationInterface|null given./'
         - '/Parameter #1 $iterator of function iterator_to_array expects Traversable, array|Traversable given./'
+        - '#[a-zA-Z0-9\\_]OrderResourcePlugin::__construct()#'
```
lsmith77: same relates to the other two constructors .. can we actually make the regexp more specific?

PR #115, opened 2018-06-20
implementing enqueue bundle

enesaktay:
made some initial steps, interested in getting comments

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,10 @@
     "suggest": {
         "portphp/excel": "To support importing excel files, use version ^1.1",
         "ext-zip": "To support writing of Excel files",
-        "portphp/csv": "To support importing csv files, use version ^1.1"
+        "portphp/csv": "To support importing csv files, use version ^1.1",
+        "enqueue/enqueue-bundle" : "To support message queuing",
+        "enqueue/redis" : "To support message queuing",
+        "predis/predis" : "To support message queuing"
     },
     "require-dev": {
         "portphp/excel": "^1.1.0",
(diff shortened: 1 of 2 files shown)


solutionDrive-Alt:
I have a feeling, that this PR can grow huge very shortly. Do you see any possibility to break the whole topic down into smaller pieces? We had already quite a few PRs that were much too big and we wanted to try to get smaller PRs in the future. (e.g. I opened one which lasted about half a year and took ages too review for others)

enesaktay:
@solutionDrive-Alt I'll break it down, good idea!

solutionDrive-Alt:
@enesaktay I think, breaking down too much is not really a thing, because reviewing and merging is way faster, but of course it should be separated into reasonable chunks, like `adding-service-xy` and within just service-xy and corresponding tests are added. PRs like this can often be reviewed and merged in minutes, without reading hours into code...

enesaktay:
@solutionDrive-Alt if we don't merge it, my other Branch-CI-Tests will fail... how are we going to handle this

stefandoorn:
I would go with an 'enqueue' branch with small PR's to it, then merge it to master if the whole solution is production worthy. As we didn't have a release for a while, it might be that users are still loading `dev-master` in their projects, which is risky in this case.

solutionDrive-Alt:
Agreed. Then we could use this Branch as the base-branch I think. And all further PRs that are enqueue related should be created against(and started from) this one.

lsmith77:
yeah .. I also agree an enqueue or actually probably `message-queue-support` branch makes sense.

lsmith77:
https://github.com/FriendsOfSylius/SyliusImportExportPlugin/tree/message-queue-support

enesaktay:
@lsmith77 so this can be merged now?

lsmith77:
done .. and also created https://github.com/FriendsOfSylius/SyliusImportExportPlugin/pull/120

PR #116, opened 2018-06-22
export send data to mq

enesaktay:


enesaktay:
@lsmith77  @stefandoorn  @solutionDrive-Alt maybe something you guys could review?

src/Command/ExportDataCommand.php
```diff
@@ -73,27 +108,71 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
 
+        return $idsToExport;
+    }
+
+    /**
+     * @param string $name
+     * @param string $file
+     * @param array $idsToExport
+     */
+    public function exportToFile(string $name, string $file, array $idsToExport): void
```
stefandoorn: Private?

src/Command/ExportDataCommand.php
```diff
@@ -73,27 +108,71 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
         foreach ($allItems as $item) {
             /** @var ResourceInterface $item */
             $idsToExport[] = $item->getId();
         }
 
+        return $idsToExport;
+    }
+
+    /**
+     * @param string $name
```
stefandoorn: Do we need these?

src/Command/ExportDataCommand.php
```diff
@@ -73,27 +108,71 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
         $service->export($idsToExport);
-        $service->finish();
+    }
 
+    /**
+     * @param string $name
+     * @param array $idsToExport
+     * @param string $exporter
+     */
+    public function exportToMq(string $name, array $idsToExport, string $exporter): void
```
stefandoorn: In here you use MQ, but sometimes also MessageQueue, I think the latter is nicer.

src/Command/ExportDataCommand.php
```diff
@@ -73,27 +108,71 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+    {
+        /** @var ResourceExporterInterface $service */
+        $service = $this->exporterRegistry->get($name);
+        $service->export($idsToExport);
+        $data = $service->getExportedData();
+
+        $mqItemWriter = new MqItemWriter(new RedisConnectionFactory());
```
stefandoorn: Factory / DI?

stefandoorn: Alternative: a different command that uses messages queues, which will only be loaded if the MQ package is loaded? Feels more 'single responsibility' to me. Each command does one thing and can use shared things (e.g. abstract class).

src/Command/ExportDataCommand.php
```diff
@@ -127,10 +208,45 @@ private function listExporters(
         $list = [];
         foreach ($exporters as $exporter => $formats) {
             // prints the exporterentity, implodes the types and outputs them in a form
+            if ($mode === '1') {
```
stefandoorn: Might use a `switch` here to improve readability.

src/Exporter/MqItemWriter.php
```diff
@@ -0,0 +1,67 @@
...
+     */
+    private $queue;
+
+    /**
+     * @var PsrConsumer
+     */
+    private $consumer;
+
+    /**
+     * @param RedisConnectionFactory $redisConnectionFactory
```
stefandoorn: Needed?

src/Exporter/MqItemWriter.php
```diff
@@ -0,0 +1,67 @@
...
+        $this->consumer = $this->redisContext->createConsumer($this->queue);
+    }
+
+    /**
+     * @param array $items
+     * @param string $entity
+     */
+    public function write(array $items, string $entity): void
+    {
+        foreach ($items as $item) {
```
stefandoorn: @lsmith77 @solutionDrive-Alt I think we should start looking into using generators rather soon :-)

src/Exporter/ResourceExporter.php
```diff
@@ -80,6 +80,25 @@ public function export(array $idsToExport): void
...
+     * @param array $idsToExport
+     *
+     * @return array
+     */
+    public function exportData(array $idsToExport): array
+    {
+        $this->pluginPool->initPlugins($idsToExport);
+        $this->writer->write($this->resourceKeys);
+
+        $dataArray = [];
```
stefandoorn: Can this have a more descriptive name?

src/Command/ExportDataCommand.php
```diff
@@ -58,13 +61,45 @@ protected function configure(): void
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
-        $exporter = $input->getArgument('exporter');
+        $mode = $input->getOption('mode');
+
+        $modes = [
+            '1' => 'Create file',
```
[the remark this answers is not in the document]
enesaktay: how about
```private const MODE_MESSAGE_QUEUE = [
        '1  ' => 'Create file',
        '2' => 'Write to message queue',
    ];``` ?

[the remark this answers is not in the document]
Comment (lsmith77) [reply src/Command/ExportDataCommand.php]: I would rather suggest to go in the direction @stefandoorn suggested. having scalar’s for your constants means you can rely on tab completion to reduce typos and when doing refactorings it is much easier to find all places to fix.

File: src/Command/ExportDataCommand.php [hunk 10]
```diff
@@ -58,13 +61,45 @@ protected function configure(): void
...
+            '2' => 'Write to message queue',
+        ];
 
-        if (empty($exporter)) {
-            $message = 'choose an exporter';
-            $this->listExporters($input, $output, $message);
+        // check if given method exists
+        if (!empty($mode) && (!array_key_exists($mode, $modes))) {
+            $message = 'choose a mode';
+            $this->listMethods($input, $output, $message, $modes);
```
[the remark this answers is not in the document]
enesaktay: it is being exited inside listMethods

src/Command/ExportDataCommand.php
```diff
@@ -58,13 +61,45 @@ protected function configure(): void
...
         $format = $input->getOption('format');
+
+        if (empty($exporter) || (empty($format) && !array_key_exists($mode, $modes))) {
+            $message = 'choose an exporter and format';
+            $this->listExporters($input, $output, $message, $mode);
+        }
+
+        /** @var RepositoryInterface $repository */
+        $repository = $this->container->get('sylius.repository.' . $exporter);
+        $allItems = $repository->findAll();
```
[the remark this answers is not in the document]
enesaktay: well allItems as the name implies has all Items but in `src/Resources/config/services_export_csv.yml`
I specify which IDs I want to be exported, which is the reason for this naming.. do you think items fits better?

src/Exporter/MqItemWriter.php
enesaktay: yes, its needed to create the context which then connects to the queue :)

src/Command/ExportDataCommand.php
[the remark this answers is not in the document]
enesaktay: this makes it more undynamic with the output of the possible modes, i'll try and find a solution in the middle 👍

src/Command/ExportDataCommand.php
enesaktay: was already like this, maybe could refactor the code in a different issue/pr?

src/Command/ExportDataCommand.php
enesaktay: I had the same thought.. I thought, if its doing the same thing except for where the data is being sent at, using the same command felt more reasonable.

enesaktay:
made some changes, open for review again

src/Command/ExportDataCommand.php
```diff
@@ -20,6 +22,10 @@
 {
     use ContainerAwareTrait;
 
+    // if a mode is added please, accordingly, add it to $modes too
+    private const MODE_CREATE_FILE = '1';
```
lsmith77: again I would use `file` and `mq` here .. but maybe its best to just have separate commands

stefandoorn: I'm +1 for separate commands wiht e.g. an abstract class under it. Separates intent and allows us even to unregister the command if no MQ in there (keeping it optional).

enesaktay: @lsmith77 @stefandoorn done 👍

src/Command/ExportDataCommand.php
```diff
@@ -4,7 +4,9 @@
 
 namespace FriendsOfSylius\SyliusImportExportPlugin\Command;
 
+use Enqueue\Redis\RedisConnectionFactory;
```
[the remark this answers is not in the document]
lsmith77: discussed this briefly with @enesaktay .. the goal is indeed to be MQ agnostic and this will hopefully happen in a future PR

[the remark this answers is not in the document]
lsmith77: but I guess this use statement can now be removed from this file?

src/Command/ExportDataCommand.php
```diff
@@ -78,22 +88,71 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
         $service->export($idsToExport);
-        $service->finish();
+    }
 
+    /**
+     * @param string $name
+     * @param array $idsToExport
+     * @param string $exporter
+     */
+    private function exportToMessageQueue(string $name, array $idsToExport, string $exporter): void
```
lsmith77: should be removed

src/Command/ExportDataToMessageQueueCommand.php
```diff
@@ -0,0 +1,170 @@
...
+        $io->listing($list);
+        exit(0);
+    }
+
+    /**
+     * @param array $allItems
+     *
+     * @return array
+     */
+    private function prepareExport(array $allItems): array
```
lsmith77: can be removed?

src/Controller/ExportDataController.php
```diff
@@ -90,6 +90,7 @@ private function exportData(Request $request, string $exporter, string $format,
         $service = $this->registry->get($name);
 
         $resources = $this->findResources($configuration, $this->findRepository($resource));
+
```
lsmith77: please revert

src/Exporter/MqItemWriter.php
lsmith77: most likely this will eventually be refactored to inject the context, since this is MQ agnostic but this can be done in another PR

src/Command/ExportDataCommand.php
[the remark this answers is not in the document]
enesaktay: it is being used at line 143 by running `new RedisConnectionFactory()`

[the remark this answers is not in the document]
enesaktay: woops my bad, forgot this was the other command 😅

src/Command/ExportDataToMessageQueueCommand.php
enesaktay: nope.. was a mistake by me, its job is to sort out only the wanted export ids

src/Command/ExportDataCommand.php
```diff
@@ -92,7 +92,6 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $service->setExportFile($file);
 
         $service->export($idsToExport);
-        $service->finish();
```
lsmith77: I guess this also needs to be reverted?

enesaktay: yep, thanks

src/Command/ExportDataToMessageQueueCommand.php
```diff
@@ -0,0 +1,170 @@
+<?php
```
stefandoorn: @enesaktay @lsmith77 Is this command sharing a lot of things with the other export command? If so, should we go abstract and share some methods / things?

PR #129, opened 2018-07-18
Add customer export 

enesaktay:


tests/Behat/Context/CliCustomersContext.php
```diff
@@ -0,0 +1,22 @@
...
+
+use Behat\Gherkin\Node\TableNode;
+use PHPUnit\Framework\Assert;
+
+class CliCustomersContext extends CliBaseContext
+{
+    /**
+     * @Then I should have at least the following customer emails in the database:
+     */
+    public function iShouldHaveAtLeastTheFollowingCustomerEmailsInTheDatabase(TableNode $customerEmails)
```
solutionDrive-Alt: does this Method belong to this PR? It seems to belong to import-Feature instead of export-feature

features/export/cli/exporting_customers_to_csv_via_cli.feature
```diff
@@ -0,0 +1,13 @@
+@managing_customers
+Feature: exporting customers to csv-file
+  In customer to have my customers exported to an external target
```
solutionDrive-Alt: The first part of this sentence sounds a bit odd.

enesaktay: Yup.. copying over the order feature and replacing order with customer might have not been the best idea 🙈

tests/Behat/Context/CliCustomersContext.php
solutionDrive-Alt: And an empty context is no option for this PR? Then the method could be moved to the PR where the Feature-Step is also introduced.

PR #137, opened 2018-07-23
Remove redundant & refine docblocks, updated ECS, require Sylius CS (closes#118)

stefandoorn:
Both Behat & Src folder done.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,13 +18,13 @@
         "predis/predis" : "To support message queuing"
     },
     "require-dev": {
-        "portphp/excel": "^1.1.0",
-        "portphp/csv": "^1.1.0",
         "behat/behat": "^3.3",
         "behat/mink": "^1.7",
         "behat/mink-browserkit-driver": "^1.3",
         "behat/mink-extension": "^2.2",
         "behat/mink-selenium2-driver": "^1.3",
+        "enqueue/enqueue-bundle": "^0.8.31",
+        "enqueue/redis": "^0.8.23",
         "friends-of-behat/context-service-extension": "^1.0",
         "friends-of-behat/cross-container-extension": "^1.0",
         "friends-of-behat/performance-extension": "^1.0",
@@ -36,12 +36,12 @@
         "phpspec/phpspec": "^3.2",
         "phpstan/phpstan-shim": "^0.9.2",
         "phpunit/phpunit": "^5.6",
+        "portphp/csv": "^1.1.0",
+        "portphp/excel": "^1.1.0",
+        "predis/predis": "^1.1",
         "se/selenium-server-standalone": "^2.52",
-        "sylius-labs/coding-standard": "^1.0",
-        "symplify/easy-coding-standard": "^2.4",
-        "enqueue/enqueue-bundle": "^0.8.31",
-        "enqueue/redis": "^0.8.23",
-        "predis/predis": "^1.1"
+        "sylius-labs/coding-standard": "^2.0",
+        "symplify/easy-coding-standard": "^4.5"
     },
     "prefer-stable": true,
     "minimum-stability": "alpha",
--- a/src/Command/ExportDataCommand.php
+++ b/src/Command/ExportDataCommand.php
@@ -25,9 +25,6 @@ final class ExportDataCommand extends Command
      */
     private $exporterRegistry;
 
-    /**
-     * @param ExporterRegistry $exporterRegistry
-     */
     public function __construct(ExporterRegistry $exporterRegistry)
     {
         $this->exporterRegistry = $exporterRegistry;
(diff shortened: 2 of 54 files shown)


src/DependencyInjection/Compiler/RegisterExporterPass.php
```diff
@@ -78,8 +71,6 @@ private function registerEventListenersForExportButton(ContainerBuilder $contain
     }
 
     /**
-     * @param ContainerBuilder $container
-     * @param string $type
```
lsmith77: looks like one more param can do away just below?

src/Exporter/ItemWriterInterface.php
```diff
@@ -6,13 +6,10 @@
...
 interface ItemWriterInterface
 {
-    /**
-     * @param string $queueName
-     */
     public function initQueue(string $queueName): void;
 
     /**
-     * @param array $items
+     * @param mixed[] $items
```
lsmith77: why `mixed[]` instead of `array` ?

src/Exporter/MqItemWriter.php
```diff
@@ -50,7 +44,7 @@ public function initQueue(string $queueName): void
     }
 
     /**
-     * @param array $items
+     * @param mixed[] $items
```
lsmith77: why `mixed[]` instead of array?

src/Exporter/ItemWriterInterface.php
stefandoorn: Mainly to be extra expressive about the fact that there's not a hard need to put either an object, integer, string, etc.

src/Exporter/MqItemWriter.php
stefandoorn: Mainly to be extra expressive about the fact that there's not a hard need to put either an object, integer, string, etc.

src/Exporter/ItemWriterInterface.php
lsmith77: but usually it will be an array if items of the same type within any given call .. as such mixed to me makes it less clear because it makes it seem as if the array usually contains elements of different types.

stefandoorn: True true. Maybe @enesaktay knows what best goes in here, I couldn't figure it out from just reading code.

src/DependencyInjection/Compiler/RegisterExporterPass.php
stefandoorn: What do you mean here?

lsmith77: `      * @param array $formats`

stefandoorn:
Updated ECS + added Sylius CS (cc @pamil). PHPStan errors are WIP.

composer.json
```diff
@@ -18,13 +18,13 @@
         "predis/predis" : "To support message queuing"
     },
     "require-dev": {
-        "portphp/excel": "^1.1.0",
```
stefandoorn: These changes are just ordering it alphabetically.

src/Exporter/ItemWriterInterface.php
enesaktay: while `$items` is currently always an array of the same entities off the DB, write could also be potentially used to write items of different entities to the message queue...
So @lsmith77 is right, as it is usually the items of the same type, i'd say it makes more sense to keep array to make it less confusing

stefandoorn: Ok, but then `ResourceInterface[]` would be better right? Else I think it's supposed to be `mixed[]` if multiple things are allowed, or maybe `object[]` is better?

enesaktay: Yeah `ResourceInterface[]` would make sense :)

stefandoorn:
Feedback processed @lsmith77 @enesaktay!

lsmith77:
so nice .. so clean :)

src/Command/ExportDataCommand.php
```diff
@@ -74,12 +71,10 @@ protected function execute(InputInterface $input, OutputInterface $output): void
...
         /** @var RepositoryInterface $repository */
         $repository = $this->container->get('sylius.repository.' . $exporter);
-        $allItems = $repository->findAll();
-        $idsToExport = [];
-        foreach ($allItems as $item) {
-            /** @var ResourceInterface $item */
-            $idsToExport[] = $item->getId();
-        }
+        $items = $repository->findAll();
+        $idsToExport = array_map(function (ResourceInterface $item) {
```
gabiudrescu: have you written this by hand or it was converted by a tool?

stefandoorn: This kind of change by hand. The docblock stuff initially also by hand, but later on used PHP-CS-Fixer to help me on it. But yeah, 95% by hand..

gabiudrescu: also, my knowledge is that array_map performs worse than a classic foreach in PHP

gabiudrescu: it would have been nice to have a foreach to array_map convert :grin:

stefandoorn: True, but probably hard to identify when it can be done :) Regarding performance - might be, but I think in this case we can prefer DX over speed.

https://stackoverflow.com/a/26527704/616307 <- not that much difference btw

gabiudrescu: for later reference: https://3v4l.org/u7Tgn#output

we're talking about a penalty of almost 60% here :disappointed: 

and I guess that this is quite a hot path in the code.

stefandoorn: Haha yeah, but this is a 500.000 items test, the whole plugin will run you out of memory anyway right now as it tries to load all entities at once instead of looping over them using generators. Something to improve when going with generators.

PR #138, opened 2018-07-25
Add a timeout function to the import-from-message-queue command

enesaktay:
The timeout makes the `import-from-message-queue` wait for a specific amount of ms for messages to arrive in the queue
 & renamed the import function as the name was misleading


src/Command/ImportDataFromMessageQueueCommand.php
```diff
@@ -38,7 +39,9 @@ protected function configure(): void
             ->setDescription('Import data from message queue.')
             ->setDefinition([
                 new InputArgument('importer', InputArgument::OPTIONAL, 'The importer to use.'),
-            ]);
+                new InputOption('timeout', null, InputOption::VALUE_OPTIONAL, 'The time in ms the importer will wait for some input.', 0),
```
stefandoorn: What if you default this to `null` and make it optional further in the code?

src/Exporter/ItemReaderInterface.php
```diff
@@ -8,7 +8,7 @@
 {
     public function initQueue(string $queueName): void;
 
-    public function readAndImport(): void;
+    public function readAndImport(int $timeout): void;
```
stefandoorn: E.g. `?int $timeout = null`?

enesaktay: good suggestion, i didn't know about the nullable types :)
is the `= null` part really needed?

solutionDrive-Alt: I would prefer to have a default-value for the input-parameter in the command (e.g. 0), because I think it's confusing if a method can be called with a parameter or not and if we default it in this method-definition, there automatically has to be a some logic to test for null, which adds complexity that is not really usefull I think. 

Any other thoughts on that?

stefandoorn: Alright, in the command seems fine to me, but I would say that the `readAndImport` method defined in the interface should have it optional. If you would default it to 0, it's also kind of optional, but `null` implies more the fact that it's optional imo.

src/Command/ImportDataFromMessageQueueCommand.php
lsmith77: why not define a short cut `t` ?

src/Exporter/ItemReaderInterface.php
lsmith77: imho in generally nullable and optional arguments should be avoided where possible .. so I think its fine like this

PR #141, opened 2018-07-25
Added phpstan-symfony & phpstan-webmozart-assert

stefandoorn:
Including fixes for new issues PHPStan discovered.

cc @enesaktay


tests/Application/app/AppKernel.php
```diff
@@ -30,20 +30,4 @@ public function registerContainerConfiguration(LoaderInterface $loader): void
     {
         $loader->load($this->getRootDir() . '/config/config.yml');
     }
-
```
stefandoorn: This is needed to have the cache folder at a predictable location for the `phpstan.neon` config.

Issue #143, opened 2018-08-16
Support for Symfony 4.1


kleinkoerkamp:
Because of symfony/stopwatch dependency is on ^3.3, the plugin is not compatible with Symfony 4.1

solutionDrive-Alt:
fixed by #145

PR #145, opened 2018-08-16
Support for Symfony 4.1

kleinkoerkamp:
Added support for Symfony 4.1. Issue #143

### DIFF
--- a/src/DependencyInjection/Compiler/RegisterExporterPass.php
+++ b/src/DependencyInjection/Compiler/RegisterExporterPass.php
@@ -23,7 +23,7 @@ final class RegisterExporterPass implements CompilerPassInterface
     public function process(ContainerBuilder $container)
     {
         $serviceId = 'sylius.exporters_registry';
-        if (!$container->has($serviceId)) {
+        if ($container->has($serviceId) == false) {
             return;
         }
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,11 @@ env:
     global:
         - SYLIUS_CACHE_DIR=$HOME/.sylius-cache
         - SYLIUS_BUILD_DIR=etc/build
+    matrix:
+        - DEPENDENCY_VERSIONS="sylius/sylius:1.2.* symfony/symfony:4.1.*"
+        - DEPENDENCY_VERSIONS="sylius/sylius:1.2.* symfony/symfony:3.4.*"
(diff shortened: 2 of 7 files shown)


stefandoorn:
I think we should also start testing this library against multiple Symfony versions and multiple Sylius versions, e.g. like done here: https://github.com/stefandoorn/sitemap-plugin/blob/master/.travis.yml#L7 & https://github.com/stefandoorn/sitemap-plugin/blob/master/.travis.yml#L21. Would you be able to add this @kleinkoerkamp ?

stefandoorn:
Also, do all services need to become public? Or only some (e.g. controllers)? It feels a bit much to me to make all of them public at once.

kleinkoerkamp:
@stefandoorn good point, i will look into that

src/DependencyInjection/Compiler/RegisterExporterPass.php
```diff
@@ -23,7 +23,7 @@
     public function process(ContainerBuilder $container)
     {
         $serviceId = 'sylius.exporters_registry';
-        if (!$container->has($serviceId)) {
+        if ($container->has($serviceId) == false) {
```
stefandoorn: Nitpicking, but this can be `===` I think.

kleinkoerkamp: `===` will give the following issue:
```
 ------ ----------------------------------------------------------------------------------- 
  Line   src/DependencyInjection/Compiler/RegisterExporterPass.php                          
 ------ ----------------------------------------------------------------------------------- 
  26     Strict comparison using === between true and false will always evaluate to false.  
 ------ ----------------------------------------------------------------------------------- 
```

`!$container->has($serviceId)` will give the following issue.
```
 ------ ----------------------------------------------------------- 
  Line   src/DependencyInjection/Compiler/RegisterExporterPass.php  
 ------ ----------------------------------------------------------- 
  26     Negated boolean is always false.                           
 ------ ----------------------------------------------------------- 
```

stefandoorn: Hm :) Ok, let's leave it for now :) Can be adjusted in another PR.

dnna:
Ping :) 
What is the status of this PR? Would really love to have this merged for my project.
If it needs further changes I'd be happy to help.

PR #147, opened 2018-09-08
Do not use internal storage for export links

dnna:
When using this bundle with [PHP-PM](https://github.com/php-pm/php-pm), the `ExportButtonGridListener` service is initialized only once, so the `$links` attribute is only populated in the first request. This PR fixes this by repopulating the links array every time the grid event occurs.


stefandoorn:
LGTM.

Issue #53, opened 2018-03-01
refactor importer to not fetch data redundantly


lsmith77:
https://github.com/FriendsOfSylius/SyliusImportExportPlugin/pull/37/files#r168944399

enesaktay:
i think the title should be "exporter" instead of "importer"? @lsmith77

PR #148, opened 2018-09-12
Make order export more efficient when using ORM

dnna:
Exporting a large number of orders is currently inefficient because the exporter has to loop through the resource set multiple times, plus it has to lazy load a number of associations.

This PR makes the export more efficient when using the ORM driver, by utilizing some Doctrine techniques to reduce the hydration overhead (read https://ocramius.github.io/blog/doctrine-orm-optimization-hydration/), as well as enable eager fetching for the exported associations.


src/Exporter/ORM/Hydrator/OrderHydrator.php
```diff
@@ -0,0 +1,87 @@
...
+
+use Doctrine\ORM\Mapping\ClassMetadata;
+use Doctrine\ORM\Query;
+use Doctrine\ORM\QueryBuilder;
+use Sylius\Component\Resource\Model\ResourceInterface;
+use Sylius\Component\Resource\Repository\RepositoryInterface;
+
+class OrderHydrator implements HydratorInterface
+{
+    protected $repository;
```
stefandoorn: Missing typehint

src/Exporter/ORM/Hydrator/OrderHydrator.php
```diff
@@ -0,0 +1,87 @@
...
+
+namespace FriendsOfSylius\SyliusImportExportPlugin\Exporter\ORM\Hydrator;
+
+use Doctrine\ORM\Mapping\ClassMetadata;
+use Doctrine\ORM\Query;
+use Doctrine\ORM\QueryBuilder;
+use Sylius\Component\Resource\Model\ResourceInterface;
+use Sylius\Component\Resource\Repository\RepositoryInterface;
+
+class OrderHydrator implements HydratorInterface
```
stefandoorn: Can we make this class final?

src/Exporter/ORM/Hydrator/OrderHydrator.php
```diff
@@ -0,0 +1,87 @@
...
+        RepositoryInterface $repository
+    ) {
+        $this->repository = $repository;
+    }
+
+    /**
+     * @param int[]|string[] $idsToExport
+     *
+     * @return ResourceInterface[]
+     */
```
stefandoorn: I think you can use inheritdoc here

src/Exporter/ORM/Hydrator/OrderHydrator.php
```diff
@@ -0,0 +1,87 @@
...
+    public function getHydratedResources(array $idsToExport): array
+    {
+        /** @var ResourceInterface[] $items */
+        if ($this->repository instanceof \Doctrine\ORM\EntityRepository) {
```
stefandoorn: Pure preference, but I think it's more readable if we return it around:

```
if (!$this->repository instanceof \Doctrine\ORM\EntityRepository) {
    return $this->repository->findBy(['id' => $idsToExport]);
}

... other things here
```

src/Exporter/ORM/Hydrator/OrderHydrator.php
```diff
@@ -0,0 +1,87 @@
...
+            $items = $this->repository->findBy(['id' => $idsToExport]);
+        }
+
+        return $items;
+    }
+
+    /**
+     * @param int[]|string[] $idsToExport
+     */
+    protected function findOrdersQb(array $idsToExport): QueryBuilder
```
stefandoorn: I think this method can be private, especially if we make the class final.

src/Exporter/ORM/Hydrator/OrderHydrator.php
```diff
@@ -0,0 +1,87 @@
...
+        // Partial hydration to make sure order items don't get lazy-loaded
+        return $this->repository->createQueryBuilder('o')
+            ->select('PARTIAL o.{id}, items')
+            ->leftJoin('o.items', 'items')
+            ->andWhere('o.id IN (:exportIds)')
+            ->setParameter('exportIds', $idsToExport)
+        ;
+    }
+
+    protected function enableEagerLoading(Query $query): Query
```
stefandoorn: Can't we make this flexible so other entities can also benefit from it? E.g. based on class metadata / relations? Or am I making it too complicated now?

src/Exporter/Plugin/ResourcePlugin.php
```diff
@@ -138,7 +138,7 @@ private function addDataForId(ResourceInterface $resource): void
      *
      * @return ResourceInterface[]
      */
-    private function findResources(array $idsToExport): array
+    protected function findResources(array $idsToExport): array
```
stefandoorn: Why does this one need to be protected now?

dnna:
Hey @stefandoorn, thanks for the comments! I will take a look into them in the next few days.

I also thought about abstracting this solution into something more general, but decided against it for 2 reasons:
- The important bits in the hydration logic are the `hydrateOrderItemsQb` and the `enableEagerLoading` functions. In order to make these generic, we would need to know which fields are actually associations to join/eager load them (so we would need to use `Reflection` or Doctrine's `ClassMetadata` objects). This would add another layer of complexity that would need testing and,
- Currently the only exportable entity that needs to utilize joins is `Order`. Other entities like `Customer` are much simpler and do not really benefit from this.

All in all I think abstracting this now is too much work for the benefits it would provide, and would be better served in a future PR, dealing only with that part, once we have more exportable entities that need to utilize joins.

stefandoorn:
@dnna Makes sense to me, let's keep it simple for now. WDYT @solutionDrive-Alt & @lsmith77?

solutionDrive-Alt:
First of all thanks to @dnna for the great work! And I'm totally ok with keeping it simpler for now.

src/Exporter/ORM/Hydrator/OrderHydrator.php
dnna: Closing this for now, too complex for this PR as discussed :)

dnna:
@solutionDrive-Alt @stefandoorn I have made the requested changes, but phpstan is failing on some unrelated files. I'm not sure if they are related to this PR.

solutionDrive-Alt:
@dnna that's strange, I'm not sure, what's happening there. The last builds used the same version of phpstan, so it should not be a version-issue. The problem seems to be only with `Commands` but I see no reason why this happens now.

@lsmith77, @stefandoorn  any idea what's happening here?

stefandoorn:
Not sure also.. @lsmith77 / @solutionDrive-Alt could you try to start a build on master to see what's the result there? If it's broken, we better fix it there, and rebase this PR on it. 

@dnna Could you rebase it anyway against master? The commit history looks a bit 'blurry' and I think all changes should be in a single commit (as they are related, not to be separated in multiple commits).

solutionDrive-Alt:
@stefandoorn @dnna I started a [new build on the master](https://travis-ci.org/FriendsOfSylius/SyliusImportExportPlugin/builds/434060462) and it failed with the same error. I will try to pin down the problem in the next days, but my guess is, that there were changes to some rules.

solutionDrive-Alt:
@dnna @stefandoorn I found the problems and fixed them in #158

solutionDrive-Alt:
@dnna the fixes are merged to `master`. Can you rebase your branch please against it? Then your build should also be green.

dnna:
I have merged the changed from master and its green now. About the commit history, won't it get squashed to a single commit by Github on the merge?

src/Controller/ExportDataController.php
```diff
@@ -101,6 +102,13 @@ private function findRepository(string $resource): RepositoryInterface
      */
     private function getResourceIds($resources): array
     {
+        if ($resources instanceof ResourceGridView &&
```
lsmith77: I prefer to move the `&&` to the beginning of the next line

ie.

```if ($resources instanceOf ResourceGridView
    && $resources->getData()->getAdapter() instanceOf DoctrineORMAdapter
) {
    ..
}
```

src/Exporter/ORM/Hydrator/OrderHydrator.php
lsmith77: I guess the default is final, since we can relax strictness, but its harder to add strictness later. and relaxing strictness is only a PR and a good argument away :)

lsmith77:
cool stuff!

lsmith77:
I guess this PR provides a pattern to fix 
https://github.com/FriendsOfSylius/SyliusImportExportPlugin/issues/53

src/Exporter/ORM/Hydrator/OrderHydrator.php
dnna: OK, good point :) I will make it final

src/Controller/ExportDataController.php
dnna: Sure :)

PR #149, opened 2018-09-13
Fix install without sylius.message_queue

spras:
If we don't need to use the message queue and don't configure `fos_sylius_import_export.message_queue.service_id` as suggested in the README, we met this error : 

```
In Reference.php line 24:
                                                                                                                                                                                                
  Argument 1 passed to Symfony\Component\DependencyInjection\Reference::__construct() must be of the type string, null given, called in /var/www/html/vendor/friendsofsylius/sylius-import-exp  
  ort-plugin/src/DependencyInjection/Compiler/MessageQueuePass.php on line 21                                                                                                                   
 ```

This PR fix this


spras:
fixe #153

Issue #160, opened 2018-12-13
documentation need to be updated?


splyy:
I tried to add this plugin to my Sylius, so far no difficulty.

However, as soon as I try to use the export (from the grid) I get a `getParameter() on null` error from `$this->container->getParameter('sylius.resources')[$exporter]);`.

And when i try form the CLI : impossible to export other entities than the default one (Customer, Order, Country).

Would there be any omissions in the documentation to use a custom export?

_(I specify that I am a junior developer)_

solutionDrive-Alt:
Hi @splyy , glad you are using this plugin. What kind of `custom export` do you have, and can you provide the `service`-definitions you created for them?

splyy:
@solutionDrive-Alt  I used the same config as the one in the documentation for exporter by replacing `foo` with my resource `brand`, as well as `bar` => `csv`

Finally, when I run `bin/console sylius:export brand file.csv --format=csv`, I find myself faced with this error ` [RuntimeException] There is no 'sylius.brand.csv' exporter.` There is no information on this subject in the documentation or on this definition.

Have I forgotten a configuration?

solutionDrive-Alt:
Based on the README I would come to the following Configuration, if I would want to add it to this repository:

in `services_csv.yml`

```yaml
sylius.exporter.brands.csv:
     class: FriendsOfSylius\SyliusImportExportPlugin\Exporter\ResourceExporter
     arguments:
        - "@sylius.exporter.csv_writer"
        - "@sylius.exporter.pluginpool.brands"
        - ["Id", "Code" ,"Enabled"] #I don't know which fields should be set here without further investigation, but here you would define the columns to export.
        - "@sylius.exporters_transformer_pool" # Optional
     tags:
        - { name: sylius.exporter, type: sylius.brand, format: csv }
```

in `services.yml`:

```yaml
sylius.exporter.pluginpool.brands:
      class: FriendsOfSylius\SyliusImportExportPlugin\Exporter\Plugin\PluginPool
      arguments:
          - ["@sylius.exporter.plugin.resource.brand"]
          - ["Id", "Code" ,"Enabled"]

sylius.exporter.plugin.resource.brand:
      class: FriendsOfSylius\SyliusImportExportPlugin\Exporter\Plugin\ResourcePlugin
      arguments:
          - "@sylius.repository.brand"
          - "@property_accessor"
          - "@doctrine.orm.entity_manager"
```

[...]

PR #161, opened 2018-12-13
Add exporter for tax categories

solutionDrive-Alt:
Missing Tests for now, started as quick test for #160. It works as expected.
Edit: Behat-Tests are now present


solutionDrive-Alt:
fixes also #162. I also realized that our behat-tests for cli-export do not test anything useful. They just ensure that the command can be executed, but the resulting files are empty (only the `header`-line is written) because we don't have any data in the db at this moment. 

I solved that for `tax_category` in this PR an will apply those changes to the other resources with the next PR.

composer.json
```diff
@@ -58,6 +59,9 @@
     "config": {
         "bin-dir": "bin"
     },
+    "conflict": {
+        "sonata-project/core-bundle": ">=3.12"
```
stefandoorn: Are all higher versions conflicting or just a single release?

tests/Behat/Context/CliBaseContext.php
```diff
@@ -127,4 +142,21 @@ public function iExportDataToSpecificFiletypeFileWithTheCliCommand($exporterType
...
+        $actualContent = file($this->exportFile);
+        foreach ($actualContent as $index => $actualRowInFile) {
+            // Check if the line in the exported file differs from the expected one
+            Assert::assertTrue($this->getDiff($expectedContent, $index, $actualRowInFile) === 0);
+        }
+    }
+
+    private function getDiff(TableNode $table, int $index, string $rowInFile): int
+    {
+        return count(array_diff($table->getRow($index), explode(';', trim($rowInFile))));
```
stefandoorn: Maybe better use [str_getcsv](http://php.net/manual/en/function.str-getcsv.php)?

tests/Behat/Context/CliTaxCategoriesContext.php
```diff
@@ -19,4 +20,20 @@ public function iShouldHaveAtLeastTheFollowingTaxCategoriesIdsInTheDatabase(Tabl
             Assert::assertNotNull($taxCategory);
         }
     }
+
+    /**
+     * @Then /^(this tax category) name is "([^"]+)"$/
+     */
+    public function thisTaxCategoryNameShouldBe(TaxCategoryInterface $taxCategory, $taxCategoryName)
```
stefandoorn: Missing void return

tests/Behat/Context/CliTaxCategoriesContext.php
```diff
@@ -19,4 +20,20 @@ public function iShouldHaveAtLeastTheFollowingTaxCategoriesIdsInTheDatabase(Tabl
...
+    {
+        $taxCategory->setName($taxCategoryName);
+    }
+
+    /**
+     * @Then /^(this tax category) description is "([^"]+)"$/
+     */
+    public function thisTaxCategoryDescriptionShouldBe(TaxCategoryInterface $taxCategory, $taxCategoryName)
```
stefandoorn: Missing void return

composer.json
solutionDrive-Alt: 3.13.4 is working fine. Should I test the versions in between?

tests/Behat/Context/CliBaseContext.php
solutionDrive-Alt: Just saw that you wrote `str_getcsv`... :-) But I think `fgetcsv` is also an improvement. :-)

tests/Behat/Context/CliBaseContext.php
```diff
@@ -69,12 +71,23 @@ class CliBaseContext implements Context
...
+     */
+    protected $objectManager;
+
     public function __construct(
         KernelInterface $kernel,
         RepositoryInterface $repository,
         ImporterRegistry $importerRegistry,
         ExporterRegistry $exporterRegistry,
         ContainerInterface $container,
+        ObjectManager $objectManager,
```
lsmith77: why was this added?

tests/Application/yarn.lock
```diff
@@ -24,6 +24,15 @@ ajv@^4.9.1:
     co "^4.6.0"
     json-stable-stringify "^1.0.1"
 
+ajv@^6.5.5:
```
lsmith77: why was this updated?

tests/Behat/Context/CountriesContext.php
```diff
@@ -6,9 +6,9 @@
 
 use Behat\Behat\Context\Context;
 use Behat\Mink\Session;
+use FriendsOfBehat\PageObjectExtension\Page\SymfonyPage;
 use PHPUnit\Framework\Assert;
 use Sylius\Behat\Context\Transform\CountryContext;
-use Sylius\Behat\Page\SymfonyPage;
```
lsmith77: should this change be done in other places too (in another PR)?

solutionDrive-Alt: Yeah, at that time I planned to fixed this for the other Pages in a follow-up PR.

tests/Behat/Context/CliBaseContext.php
solutionDrive-Alt: I think I forgot to remove this, when I was cleaning up this PR. I also realized at this time, that our behat tests for the export aren't very useful at the moment, because they don't test more than command-execution. #163

tests/Application/yarn.lock
solutionDrive-Alt: Honestly, I don't recall updating this. Maybe also related to #163 . I will check this later today or this weekend.

solutionDrive-Alt: it looks like it is related to the update of sylius from version `1.2.8` to `1.2.12`, which I have done. Shall I revert this (update of sylius)?

tests/Behat/Context/CliBaseContext.php
solutionDrive-Alt: I removed this with and all injections into the services.

tests/Behat/Context/CountriesContext.php
solutionDrive-Alt: actually this is the only Context which extends SymfonyPage, and I'm not sure if this is necessary at all, since it is a `Context` and not a `PageObject`. Every other context is using `ResourceIndexPageInterface` which extends the `\Sylius\Behat\Page\Admin\Crud\IndexPage`. And this already uses the correct PageObject of the `FriendsOfBehat\PageObjectExtension`. Problably this was also the reason for the `Sylius`-update I made.

tests/Behat/Context/CliBaseContext.php
solutionDrive-Alt: @stefandoorn is this also ok, for you or is there a reason why I should change it to `str_getcsv`?

stefandoorn: Yeah if it works :-) I try to use native functions for these things instead of rolling it myself, but I think this does the job :-)

tests/Behat/Context/CountriesContext.php
stefandoorn: Fixed in #168

lsmith77:
please rebase on master

solutionDrive-Alt:
On it... :-)

solutionDrive-Alt:
@stefandoorn I think I addressed all feedback.

solutionDrive-Alt:
@lsmith77 @stefandoorn I will merge this since both of you approved.

PR #169, opened 2019-02-18
Switch excel lib

lsmith77:
https://github.com/portphp/portphp/issues/87

Note with this lib we could start supporting several other output formats:
https://phpspreadsheet.readthedocs.io/en/develop/


solutionDrive-Alt:
nice!

Issue #171, opened 2019-02-22
CSV delimiter different in import and export files


oallain:
Hi,

I think, the fact of the delimiter of the CSV file is semicolon (;) for exports and comma (,) for imports is not logical.

As a result, we can not import a file we just exported :(

I propose to use the comma (,) delimiter for exports.

What do you think ?

stefandoorn:
My preference would be ; for both, but I'm not sure what tools like Excel export generally. Probably best to stay as close as possible to these, as that makes working with imports easier.

oallain:
> Comma separated values text files (.csv), in which the comma character (,) typically separates each field of text.

Read at https://support.office.com/en-us/article/import-or-export-text-txt-or-csv-files-5250ac4c-663c-47ce-937b-339e391393ba

I'm on Linux so and can't make test now :disappointed:

stefandoorn:
Then it makes sense to stick to a comma, cc @lsmith77 @solutionDrive-Alt.

Would you be able to open a PR with the suggested changes?

solutionDrive-Alt:
I'm totally fine with `comma` as separator.

lsmith77:
yeah, lets go with `comma` by default

oallain:
Hello,

I do PR #172 for fix that :)

solutionDrive-Alt:
fixed by #172

oallain:
Thank you ! :tada:

PR #172, opened 2019-02-25
set comma to default csv delimiter

oallain:
Fix issue #171


stefandoorn:
LGTM. I think we should be able to support multiple import delimiters btw, but that's another issue than this :-) Thanks :)

oallain:
@stefandoorn yes, we can, by config, but like you say, that's for an other PR :+1:

Issue #91, opened 2018-03-30
Define controllers as services


stefandoorn:
And not extend base controller, inject services instead of fetching them inside the methods.

PR #173, opened 2019-02-27
set controller as service

oallain:
Feature for "Define controllers as services" #91

### DIFF
--- a/README.md
+++ b/README.md
@@ -313,7 +313,7 @@ app_export_data_foo:
             filterable: true
             grid: sylius_admin_foo # Name of defined grid here
 ```
-and the associate controller
+And add the associated controller service definition to your services
 
 ```yaml
 sylius.controller.export_data_foo:
@@ -389,7 +389,7 @@ app_export_data_country:
             grid: sylius_admin_country
 ```   
 
-and the associate controller
+And add the associated controller service definition to your services
 
 ```yaml
 sylius.controller.export_data_country:
--- a/README.md
+++ b/README.md
@@ -308,11 +308,26 @@ app_export_data_foo:
     methods: [GET]
     defaults:
         resource: sylius.foo
-        _controller: sylius.controller.export_data:exportAction
+        _controller: sylius.controller.export_data_foo:exportAction
         _sylius:
             filterable: true
             grid: sylius_admin_foo # Name of defined grid here
 ```
+and the associate controller
+
+```yaml
+sylius.controller.export_data_foo:
+    public: true
+    class: FriendsOfSylius\SyliusImportExportPlugin\Controller\ExportDataController
+    arguments:
+        - "@sylius.exporters_registry"
+        - "@sylius.resource_controller.request_configuration_factory"
+        - "@sylius.resource_controller.resources_collection_provider"
+        - "@sylius.repository.foo"
+        - "%sylius.resources%"
+    tags: ['controller.service_arguments']
+```
+
 
 In case you don't add it, the UI exporters will still function. They will simply load all data of that resource for the export (similar as CLI).
 
@@ -368,12 +383,26 @@ app_export_data_country:
     methods: [GET]
     defaults:
         resource: sylius.country
-        _controller: sylius.controller.export_data:exportAction
+        _controller: sylius.controller.export_data_contry:exportAction
         _sylius:
             filterable: true
             grid: sylius_admin_country
 ```   
-   
+
+and the associate controller
+
+```yaml
+sylius.controller.export_data_country:
+    public: true
+    class: FriendsOfSylius\SyliusImportExportPlugin\Controller\ExportDataController
+    arguments:
+        - "@sylius.exporters_registry"
+        - "@sylius.resource_controller.request_configuration_factory"
+        - "@sylius.resource_controller.resources_collection_provider"
+        - "@sylius.repository.country"
+        - "%sylius.resources%"
+    tags: ['controller.service_arguments']
+```
 
 ### PluginPool
 The idea behind the plugin pool is, to be able to have different kind of plugins, which could possibly
--- a/src/Controller/ExportDataController.php
+++ b/src/Controller/ExportDataController.php
@@ -19,18 +19,18 @@ use Sylius\Component\Registry\ServiceRegistryInterface;
 use Sylius\Component\Resource\Metadata\Metadata;
 use Sylius\Component\Resource\Model\ResourceInterface;
 use Sylius\Component\Resource\Repository\RepositoryInterface;
-use Symfony\Component\DependencyInjection\ParameterBag\ParameterBagInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpFoundation\ResponseHeaderBag;
 
 final class ExportDataController
 {
+    /** @var \Symfony\Component\DependencyInjection\ContainerInterface */
+    private $container;
+
     /** @var EntityManager */
     private $entityManager;
-
-    /** @var ParameterBagInterface */
-    private $parameterBag;
 
     /** @var ServiceRegistryInterface */
     private $registry;
@@ -45,14 +45,14 @@ final class ExportDataController
         ServiceRegistryInterface $registry,
         RequestConfigurationFactoryInterface $requestConfigurationFactory,
         ResourcesCollectionProviderInterface $resourcesCollectionProvider,
-        ParameterBagInterface $parameterBag,
+        ContainerInterface $container,
         EntityManager $entityManager
     ) {
         $this->registry = $registry;
         $this->requestConfigurationFactory = $requestConfigurationFactory;
         $this->resourcesCollectionProvider = $resourcesCollectionProvider;
-        $this->parameterBag = $parameterBag;
         $this->entityManager = $entityManager;
+        $this->container = $container;
     }
 
     public function exportAction(Request $request, string $resource, string $format): Response
@@ -66,7 +66,7 @@ final class ExportDataController
     {
         [$applicationName, $resource] = explode('.', $exporter);
         $metadata = Metadata::fromAliasAndConfiguration($exporter,
-            $this->parameterBag->get('sylius.resources')[$exporter]);
+            $this->container->getParameter('sylius.resources')[$exporter]);
         $configuration = $this->requestConfigurationFactory->create($metadata, $request);
 
         $name = ExporterRegistry::buildServiceName($exporter, $format);
@@ -93,10 +93,10 @@ final class ExportDataController
     {
         $parameter = \sprintf('sylius.model.%s.class', $resource);
 
-        if (!$this->parameterBag->has($parameter)) {
+        if (!$this->container->hasParameter($parameter)) {
             throw new \Exception(sprintf("Parameter '%s' is not defined", $parameter));
         }
-        $entityName = $this->parameterBag->get($parameter);
+        $entityName = $this->container->getParameter($parameter);
 
         return new EntityRepository($this->entityManager, new ClassMetadata($entityName));
     }
--- a/src/Controller/ExportDataController.php
+++ b/src/Controller/ExportDataController.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 namespace FriendsOfSylius\SyliusImportExportPlugin\Controller;
 
-use Doctrine\ORM\EntityManager;
 use FriendsOfSylius\SyliusImportExportPlugin\Exporter\ExporterRegistry;
 use FriendsOfSylius\SyliusImportExportPlugin\Exporter\ResourceExporterInterface;
 use Pagerfanta\Adapter\DoctrineORMAdapter;
@@ -29,9 +28,6 @@ final class ExportDataController
     /** @var RepositoryInterface */
     private $repository;
 
-    /** @var EntityManager */
-    private $entityManager;
-
     /** @var ServiceRegistryInterface */
     private $registry;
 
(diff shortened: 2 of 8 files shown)


oallain:
Hello @stefandoorn 

Does this PR match your expectation for issue #91?

lsmith77:
thank you for working on this!

I would want to try to avoid to inject the container, if at all we should inject a service locator. however I think it might be even better if we inject a service that has the responsibility to determine the entity name and that service uses a service locator.

https://symfony.com/doc/current/service_container/service_subscribers_locators.html#defining-a-service-locator

oallain:
Hello @lsmith77,

I would like too, but the ParameterBag only became a service from Symfony 4.1 (see 
https://symfony.com/blog/new-in-symfony-4-1-getting-container-parameters-as-a-service), so i haven't any other simple solution to get parameter dynamically.

Do you have an idea to get workaround this?

lsmith77:
@oallain the workaround I am proposing is to define a service that implements `ParameterBagInterface` of our own that does have the container injected to fetch parameters. this way we have at least separated this "uglyness" from the controller.

since `Symfony\Component\DependencyInjection\ParameterBag\ParameterBagInterface` has been around a long time, we can easily type hint it and then have some BC logic inside our DI FOSSyliusImportExportExtension to determine if we need to inject our BC service or can rely on Symfony autowiring.

does that make sense?

oallain:
@lsmith77 job done :+1:

src/Resources/config/services.yml
```diff
@@ -1,13 +1,20 @@
 services:
+    # Parameter Bag
```
lsmith77: @solutionDrive-Alt / @stefandoorn wdyt should we try to use the core feature for 4.1 https://symfony.com/blog/new-in-symfony-4-1-getting-container-parameters-as-a-service ?

this would require some extra work in the DI Extension to determine the Symfony version to decide if autowiring will already inject the parameter bag automatically or not.

src/Controller/ExportDataController.php
```diff
@@ -84,15 +101,14 @@ private function exportData(Request $request, string $exporter, string $format,
...
 
-        /** @var \Sylius\Component\Resource\Repository\RepositoryInterface $repository */
-        $repository = $this->get($repositoryName);
+        if (!$this->parameterBag->has($parameter)) {
+            throw new \Exception(sprintf("Parameter '%s' is not defined", $parameter));
+        }
+        $entityName = $this->parameterBag->get($parameter);
 
-        return $repository;
+        return new EntityRepository($this->entityManager, new ClassMetadata($entityName));
```
lsmith77: I think we should rather use `$this->entityManager->getRepository($entityName);` here

see https://github.com/doctrine/persistence/blob/master/lib/Doctrine/Persistence/ObjectManager.php#L110

that being said previously we fetched the repository as a service which of course is also an extension point, which we now side-step. I wonder if Sylius provides a factory service that receives the `$entityName` and returns the given `EntityRepository` instance?

lsmith77: posted a question about this in #general on Sylius slack

src/Resources/config/services.yml
solutionDrive-Alt: Sounds interesting. But probably something for a different PR.

src/Controller/ExportDataController.php
lsmith77: maybe using `$em->getRepository()` can be considered safe .. confirming here https://github.com/Sylius/Sylius/issues/873

lsmith77: @oallain if you have time .. maybe you can test if one really gets the same instance via `$em->getRepository()` as via `$container->get()` .. because if we do, then imho we can wrap this PR up.

src/DependencyInjection/ParameterBag/ParameterBag.php
```diff
@@ -0,0 +1,43 @@
+<?php
```
lsmith77: maybe add a comment here that this is a BC solution only needed while we support versions below 4.1 and add a link to https://symfony.com/blog/new-in-symfony-4-1-getting-container-parameters-as-a-service

src/Controller/ExportDataController.php
oallain: Hi @lsmith77 ,

Work :

- `$this->entityManager->getRepository(\Sylius\Component\Core\Model\Order::class)`

Don't work :

- `$this->entityManager->getRepository('Order')`
- `$this->entityManager->getRepository('sylius.model.order.class')`

But, I just have the class name like "country", "order" ....

So, and don't know how to make it better :(

src/DependencyInjection/ParameterBag/ParameterBag.php
oallain: :+1:

src/Controller/ExportDataController.php
lsmith77:
```
pjedrzejewski   [24 minutes ago]
I don’t think there is other way now… except making a “repository locator” which would be a container too but just smaller


pjedrzejewski   [24 minutes ago]
the genericness of this controller requires a container of some sort imo

pjedrzejewski   [23 minutes ago]
that’s why in the ResourceBundle we generate controller per resource, where we inject the right repository

lsmith   [20 minutes ago]
ok .. I guess we could do the same .. would of course increase number of routes and services .. but I guess the overhead is small .. and as an upside it adds an extension point by resource
```

@stefandoorn / @solutionDrive-Alt wdyt? should we provide a controller per resource?

lsmith77: https://sylius-devs.slack.com/archives/C3EGDG9LY/p1552896248352900

solutionDrive-Alt: I think when the importers / exporters get more sophisticated, we won't be able to handle it in that generic way anyways. On the other hand we will have to be carefull, that it don't evolves into a mess, where every Export/Import-Controller does the same thing differently.

oallain: Hello @lsmith77 and @solutionDrive-Alt ,

What do you want to do?
We validate like this, or do we do otherwise?
 
Thanks

lsmith77: so imho we should create separate services for each route allowing us to inject the specific repository instance rather than having to fetch it inside the controller.

so @oallain if you are motivated, we would appreciate it if you could do the necessary adjustments. thank for all your work!

oallain: Ok, i'll try to work on it this week.

src/Controller/ExportDataController.php
```diff
@@ -35,11 +44,17 @@ final class ExportDataController extends Controller
     public function __construct(
         ServiceRegistryInterface $registry,
         RequestConfigurationFactoryInterface $requestConfigurationFactory,
-        ResourcesCollectionProviderInterface $resourcesCollectionProvider
+        ResourcesCollectionProviderInterface $resourcesCollectionProvider,
+        RepositoryInterface $repository,
+        EntityManager $entityManager,
```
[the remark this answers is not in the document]
oallain: Yes, sorry

src/Resources/config/services.yml
```diff
@@ -1,21 +1,56 @@
...
+
+    sylius.controller.export_data_country:
+        public: true
+        class: FriendsOfSylius\SyliusImportExportPlugin\Controller\ExportDataController
+        arguments:
+            - "@sylius.exporters_registry"
+            - "@sylius.resource_controller.request_configuration_factory"
+            - "@sylius.resource_controller.resources_collection_provider"
+            - "@sylius.repository.country"
+            - "@doctrine.orm.entity_manager"
```
lsmith77: entity manager also needs to be removed from all the relevant services in this file

oallain: Yes, i was sleeping when Travis get result ;)

oallain: Fixed :)

lsmith77: and green we are!

oallain: Yeah !

lsmith77:
one last thing would be adding something to the README.md to remind users to add those controller services as needed. while I realized that I also remembered that we could make use of service inheritance to make it easier to define those resource specific services https://symfony.com/doc/current/service_container/parent_services.html

that being said, we could just as well make this a follow up ticket.

lsmith77:
@stefandoorn / @solutionDrive-Alt can you review / merge ?

As noted above, from the code it looks good to me .. but I didn't run any manual tests to confirm if things are working.

README.md
```diff
@@ -308,11 +308,26 @@ app_export_data_foo:
     methods: [GET]
     defaults:
         resource: sylius.foo
-        _controller: sylius.controller.export_data:exportAction
+        _controller: sylius.controller.export_data_foo:exportAction
         _sylius:
             filterable: true
             grid: sylius_admin_foo # Name of defined grid here
 ```
+and the associate controller
```
stefandoorn: I would change in something like: `And add the associated controller service definition to your services`.

README.md
```diff
@@ -368,12 +383,26 @@ app_export_data_country:
...
         resource: sylius.country
-        _controller: sylius.controller.export_data:exportAction
+        _controller: sylius.controller.export_data_contry:exportAction
         _sylius:
             filterable: true
             grid: sylius_admin_country
 ```   
-   
+
+and the associate controller
```
stefandoorn: Same here

src/Controller/ExportDataController.php
```diff
@@ -16,13 +16,18 @@
...
-use Symfony\Bundle\FrameworkBundle\Controller\Controller;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpFoundation\ResponseHeaderBag;
 
-final class ExportDataController extends Controller
+final class ExportDataController
 {
+    /** @var array */
+    private $syliusResources;
```
stefandoorn: What about renaming to `$resources`? We are in the context of Sylius.

src/DependencyInjection/ParameterBag/ParameterBag.php
```diff
@@ -0,0 +1,50 @@
...
+
+declare(strict_types=1);
+
+namespace FriendsOfSylius\SyliusImportExportPlugin\DependencyInjection\ParameterBag;
+
+use Symfony\Component\DependencyInjection\Container;
+use Symfony\Component\DependencyInjection\ParameterBag\FrozenParameterBag;
+use Symfony\Component\DependencyInjection\ParameterBag\ParameterBagInterface;
+
+class ParameterBag extends FrozenParameterBag implements ParameterBagInterface
```
stefandoorn: Shall we make this one final?

src/DependencyInjection/ParameterBag/ParameterBag.php
```diff
@@ -0,0 +1,50 @@
+<?php
+
+/*
+ * This file is need while SyliusImportExportPlugin support Symfony version < 4.1
```
stefandoorn: I would reword a bit here:

```
This is file is required as long as SyliusImportExportPlugin supports Symfony versions lower than 4.1. Before this version, ParameterBag wasn't a service available.

See: https://symfony.com/blog/new-in-symfony-4-1-getting-container-parameters-as-a-service
```

src/Resources/config/routing.yml
```diff
@@ -7,31 +7,31 @@ app_import_data:
 
 ## Export routes for specific settings
 app_export_data_country:
-    path: /export/sylius.country/{format}
+    path: /export/{resource}/{format}
```
stefandoorn: Is it handy to have `{resource}` here, while in the `_controller` we 'nail' it to the `export_data_country` controller? I think then we should be explicit in both.

src/Resources/config/routing.yml
```diff
@@ -7,31 +7,31 @@ app_import_data:
...
         resource: sylius.country
-        _controller: sylius.controller.export_data:exportAction
+        _controller: sylius.controller.export_data_country:exportAction
         _sylius:
             filterable: true
             grid: sylius_admin_country
 
 app_export_data_order:
-    path: /export/sylius.order/{format}
+    path: /export/{resource}/{format}
```
stefandoorn: Same here

src/Resources/config/routing.yml
```diff
@@ -7,31 +7,31 @@ app_import_data:
...
         resource: sylius.order
-        _controller: sylius.controller.export_data:exportAction
+        _controller: sylius.controller.export_data_order:exportAction
         _sylius:
             filterable: true
             grid: sylius_admin_order
 
 app_export_data_customer:
-    path: /export/sylius.customer/{format}
+    path: /export/{resource}/{format}
```
stefandoorn: Same here

README.md
oallain: fixed

src/Resources/config/routing.yml
oallain: Yes, it's more explicit ;)

stefandoorn: @lsmith77 @solutionDrive-Alt WDYT?

stefandoorn: Basically what I meant is, shouldn't we just keep `sylius.country` there? It's not variable, as it's tied to the country controller specific.

lsmith77: while the service specific, the controller code is generic. so in that sense its nice to have easy access to the "context" of which resource is being handled by this specific instance of the generic controller. that being said, when customizing the controller one has different ways to get this info:

1) by introspecting the repository instance
2) by adding another service parameter

I am unsure if these approaches are less/more elegant.
One use case that cannot be covered by them would be event listeners that get triggered before the controller is called. If those are legitimate I am not sure.

So to conclude, we could remove this placeholder for now simply because we do not yet have a concrete use case (unless someone has one). That being said, leaving it as is, imho is also not very risky. As such I am fine either or.

stefandoorn: Then I propose to leave it as in the PR for now, as I think later on we might be able to do something neat with registration based on tags instead of each controller separate.

lsmith77:
wow thanks so much @oallain !

oallain:
Thanks too ! :tada:

PR #174, opened 2019-03-04
travis add sylius 1.3

oallain:
Improve Travis for add Sylius 1.3 :tada:


solutionDrive-Alt:
Nice, maybe we should also add Sylius 1.4? relates to #178

.travis.yml
```diff
@@ -18,11 +14,37 @@ env:
...
+
+matrix:
+    include:
+        -
+            php: 7.3
+            env:
+                - DEPENDENCY_VERSIONS="sylius/sylius:1.3.* symfony/symfony:4.1.*"
+        -
+            php: 7.2
+            env:
```
solutionDrive-Alt: I think we should build `sylius:1.3` and `sylius:1.4` also in the other versions of php, which are supported by the sylius-versions.
See https://github.com/Sylius/Sylius/blob/1.4/composer.json#L22

oallain: Hi, @solutionDrive-Alt 

I add versions for Sylius 1.3 :smile: 

But Sylius 1.4 need more work (on an other PR ?), i tryed it here https://travis-ci.org/FriendsOfSylius/SyliusImportExportPlugin/jobs/507249837

solutionDrive-Alt: yeah I also ran into strange problems with 1.4

PR #175, opened 2019-03-11
Prevent exception when no exporter name supplied

stefandoorn:


src/Command/ExportDataCommand.php
```diff
@@ -60,7 +60,9 @@ protected function execute(InputInterface $input, OutputInterface $output): void
 
         if (empty($exporter)) {
             $this->listExporters($input, $output);
+            return;
```
oallain: 1 new line before `return` instruction.

src/Command/ExportDataCommand.php
```diff
@@ -60,7 +60,9 @@ protected function execute(InputInterface $input, OutputInterface $output): void
...
         }
+
```
oallain: You have space in this line.

PR #180, opened 2019-03-25
Use field names instead of column names

pamil:
We came across an issue caused by using column names instead of field names while exporting taxons. The taxon has the `right` field, which is mapped as `tree_right` column. This makes the property accessor trying to get `treeRight` property or call `getTreeRight` method which does not exist and the export crashes.


oallain:
Hi @pamil ,

I saw you have same error like me, on travis.
For fix-it i have forced twig version :
https://github.com/FriendsOfSylius/SyliusImportExportPlugin/pull/187/files#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780R9

Regards

PR #182, opened 2019-03-25
Call flush when JsonResourceImporter finishes its job

pamil:


oallain:
Hi @pamil ,

I saw you have same error like me, on travis.
For fix-it i have forced twig version :
https://github.com/FriendsOfSylius/SyliusImportExportPlugin/pull/187/files#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780R9

Regards

lsmith77:
@pamil I think you lost part of your changes here .. ?

oallain:
@lsmith77 @pamil 

I think it's due of my fixe in d43be8848b73924cc79403991b3f8450R47 
Behat test had an error, so I corrected it ;)

pamil:
Yeah, it looks like the majority of this PR was done in #187. Still, it's a nice spec fix, since `batchSize` is `int` and `false` is passed in phpSpec to the class constructor (which is called in a script without strict types enabled so it's converted to 0).

PR #187, opened 2019-03-25
fix batching logic for processors

### LINKED ISSUE #186: Default processors calls `RepositoryInterface::add` for every object
Default processors calls `RepositoryInterface::add` for every object which is inefficient, as every `add` call triggers a flush on entity manager and breaks batching logic. Injecting a manager there and persisting them instead of adding to the repository will solve this issue.

oallain:
Fixes #186 

I put `batch_size` by default to 1, for flush after each item.

### DIFF
--- a/spec/Processor/ResourceProcessorSpec.php
+++ b/spec/Processor/ResourceProcessorSpec.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace spec\FriendsOfSylius\SyliusImportExportPlugin\Processor;
 
+use Doctrine\ORM\EntityManagerInterface;
 use FriendsOfSylius\SyliusImportExportPlugin\Exception\AccessorNotFoundException;
 use FriendsOfSylius\SyliusImportExportPlugin\Processor\MetadataValidatorInterface;
 use FriendsOfSylius\SyliusImportExportPlugin\Processor\ResourceProcessor;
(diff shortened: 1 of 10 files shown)


spec/Processor/ResourceProcessorSpec.php
```diff
@@ -96,16 +100,18 @@ function it_can_process_an_array_of_country_data(
             ->shouldBeCalled();
 
         $this->process($dataset);
+        $entityManager->flush();
```
pamil: I think there's `should*` missing here.

src/DependencyInjection/Configuration.php
```diff
@@ -23,7 +23,7 @@ public function getConfigTreeBuilder(): TreeBuilder
                     ->addDefaultsIfNotSet()
                     ->children()
                         ->booleanNode('web_ui')->defaultTrue()->end()
-                        ->integerNode('batch_size')->defaultValue(0)->end()
+                        ->integerNode('batch_size')->defaultValue(1)->end()
```
pamil: It's okay for now, but I'd think it over whether we want to have one as default, it's not the most performant setting.

oallain: I put this value but I think 100 would be better. What do you think ?

oallain:
Waiting for https://github.com/FriendsOfSylius/SyliusImportExportPlugin/pull/188

src/DependencyInjection/Configuration.php
stefandoorn: I think 100 is better than 1 indeed.

stefandoorn:
Could you rebase on master (and remove both Twig commits)?

src/DependencyInjection/Configuration.php
lsmith77: agreed

oallain:
Hello, i had :

- rebase on master
- set default batch_size to 100
- fix batching logic for JsonResourceImporter too
- remove Twig commit

Regards :smile:

oallain:
Thank you for merged :smile:

PR #188, opened 2019-03-26
Remove composer.lock and support for Sylius 1.0

pamil:
It is a library so it's not really needed here.

Sylius 1.0 build is failing, but I'd recommend to drop that version since it's not supported with neither bugfixes nor security fixes anymore.


oallain:
I think too,
If we merge this PR i remove my sad commit like this https://github.com/FriendsOfSylius/SyliusImportExportPlugin/pull/187/commits/3fe68223a781c22e0e25f8a9bdce61f94c8e4c8d :smile:

PR #193, opened 2019-05-09
update for Sylius 1.4 and 1.5

oallain:
Add support of Sylius 1.4 and 1.5.


stefandoorn:
I think something is going wrong here, as I see a lot other changes also, e.g. in README / controllers.

oallain:
Yeah sorry @stefandoorn, i make some tests and i make an error ^^ It's fixed now :smile:

composer.json
```diff
@@ -4,8 +4,8 @@
     "description": "import / export plugin for Sylius.",
     "license": "MIT",
     "require": {
-        "php": "^7.1",
-        "sylius/sylius": "^1.1",
+        "php": "^7.2",
+        "sylius/sylius": "~1.4.0",
```
pamil:
```suggestion
        "sylius/sylius": "^1.4",
```

phpunit.xml.dist
```diff
@@ -4,11 +4,13 @@
          xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/5.6/phpunit.xsd"
          backupGlobals="false"
          colors="true"
-         bootstrap="vendor/autoload.php">
+         bootstrap="tests/Application/config/bootstrap.php">
     <testsuites>
         <testsuite name="SyliusImportExportPlugin Test Suite">
             <directory>tests</directory>
         </testsuite>
+        <env name="APP_ENV" value="test"/>
```
pamil:
```suggestion
        <server name="APP_ENV" value="test"/>
```

Same below. Reference: https://github.com/symfony/recipes/pull/561

.travis.yml
```diff
@@ -12,42 +12,30 @@ cache:
...
-            php: 7.2
-            env:
-                - DEPENDENCY_VERSIONS="sylius/sylius:1.2.* symfony/symfony:3.4.*"
-        -
-            php: 7.1
+            php: 7.3
             env:
-                - DEPENDENCY_VERSIONS="sylius/sylius:1.2.* symfony/symfony:3.4.*"
-        # Sylius 1.1
+                - DEPENDENCY_VERSIONS="sylius/sylius:1.4.* symfony/symfony:4.2.*"
```
pamil: I'd recommend the following matrix:

Sylius 1.4, PHP 7.2, Symfony 4.2
Sylius 1.4, PHP 7.3, Symfony 3.4
Sylius 1.5, PHP 7.2, Symfony 4.2
Sylius 1.5, PHP 7.3, Symfony 3.4

This way we test every possible Sylius version with two versions of PHP and Symfony.

oallain: @pamil i can't do it, because Sylius 1.4 require symfony/dotenv 4.2.
So i had an error as you can see https://travis-ci.org/FriendsOfSylius/SyliusImportExportPlugin/jobs/532859035#L521

Have you a solution ?

tests/Behat/Resources/services.xml
```diff
@@ -2,89 +2,79 @@
 
 <container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
     <parameters>
-        <parameter key="behat.files_directory">%__symfony__.kernel.project_dir%/tests/Behat/Resources/fixtures</parameter>
+        <parameter key="behat.files_directory">tests/Behat/Resources/fixtures</parameter>
     </parameters>
     <services>
+        <defaults public="true" />
```
oallain:
```suggestion
        <defaults public="true" />

```

composer.json
```diff
@@ -4,8 +4,8 @@
     "description": "import / export plugin for Sylius.",
     "license": "MIT",
     "require": {
-        "php": "^7.1",
-        "sylius/sylius": "^1.1",
+        "php": "^7.2",
+        "sylius/sylius": "^1.4",
```
stefandoorn: Do we really have to up this requirement, as in, won't it work at all at ^1.1? I'm fine not testing against it, but I'm not sure if we should restrict it that hard.

oallain: Yes it's works on Sylius ^1.1.
But Behat tests have breaking changes since Sylius 1.4 (like remove `__behat__.` prefix on services).  

For exemple, if i launch test on Sylius 1.3 i get that :

```The service "test.behat.page.country_index_page" has a dependency on a non-existent parameter "__behat__.mink.parameters".```

stefandoorn: Ok. I'm not sure if that should be an issue though, as I'm not running plugin Behat from my projects actually.

WDYT about keeping it at `^1.1` @pamil @lsmith77 @solutionDrive-Alt ?

lsmith77: since I do not have a project with Sylius .. for me personally its fine to bump up.

solutionDrive-Alt: I'm also fine with that.

PR #194, opened 2019-05-21
fix routing for custom export

oallain:
If we use custom exporter this "Backup route" catch page call and make an fatal error.

So i fixed by using spéficic route for each export.

This not a BC, because app_export_data_*  respect the format used and described in the doc.

### DIFF
--- a/spec/DependencyInjection/Compiler/RegisterExporterPassSpec.php
+++ b/spec/DependencyInjection/Compiler/RegisterExporterPassSpec.php
@@ -29,7 +29,8 @@ class RegisterExporterPassSpec extends ObjectBehavior
         Definition $exporterRegistry,
         Definition $blockEventDefinition
     ) {
-        $exporterType = 'csv';
+        $exporterType = 'country';
+        $exporterFormat = 'csv';
 
         $container->has('sylius.exporters_registry')->willReturn(true);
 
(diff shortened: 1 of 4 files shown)


stefandoorn:
Build is breaking, seems there are resources without a dot in it..

oallain:
Yes, i had fixed it :smile: 

Reading the code, I also realized that the spec/DependencyInjection/Compiler/RegisterExporterPassSpec.php, was not correct.

src/Listener/ExportButtonGridListener.php
```diff
@@ -89,10 +89,17 @@ private function addLink(string $format): array
             $parameters['criteria'] = $currentRequest->query->get('criteria');
         }
 
+        $ressource = $this->resource;
+
+        $explode = \explode('.', $this->resource);
```
lsmith77: imho we should not prefix core functions with `\`

oallain: Yes, you are right.
It is not part of the list of functions optimized by opcode ;)
See https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php#L346

oallain:
```suggestion
        $explode = explode('.', $this->resource);
```

spec/DependencyInjection/Compiler/RegisterExporterPassSpec.php
```diff
@@ -39,29 +40,30 @@ function it_processes_the_exporter_services(
...
         ]);
 
         $container->getParameter('sylius.exporter.web_ui')->willReturn(true);
-        $container->has('app.grid_event_listener.admin.crud_csv_exporter_format_export')->willReturn(false);
+        $container->has('app.grid_event_listener.admin.crud_' . $exporterType . '_' . $exporterFormat . '_export')->willReturn(false);
+        $container->has('sylius.controller.export_data_' . $exporterType)->willReturn(false);
 
         $container->register(
-            'app.grid_event_listener.admin.crud_csv_exporter_format_export',
+            'app.grid_event_listener.admin.crud_' . $exporterType . '_' . $exporterFormat . '_export',
```
bitbager: You can use \sprintf('app.grid_event_listener.admin.crud_%s_%s_export', $exporterType, $exporterFormat) as a shorthand :)

spec/DependencyInjection/Compiler/RegisterExporterPassSpec.php
```diff
@@ -39,29 +40,30 @@ function it_processes_the_exporter_services(
...
-        $blockEventDefinition->addArgument('csv')->willReturn($blockEventDefinition);
+        $blockEventDefinition->addArgument($exporterType)->willReturn($blockEventDefinition);
+        $blockEventDefinition->addArgument([$exporterFormat])->willReturn($blockEventDefinition);
         $blockEventDefinition->addMethodCall('setRequest', Argument::that(function ($input) {
             return is_array($input);
         }))->willReturn($blockEventDefinition);
         $blockEventDefinition->addTag('kernel.event_listener',
             [
-                'event' => 'sylius.grid.admin_csv',
+                'event' => 'sylius.grid.admin_' . $exporterType,
```
bitbager: \sprintf('sylius.gird.admin_%s', $exporterType) might work here as well

spec/DependencyInjection/Compiler/RegisterExporterPassSpec.php
oallain: Yes, of course.

oallain:
Hello @lsmith77 @stefandoorn @bitbager @solutionDrive-Alt 

This pull request is blocking for me ;)
Do you thinks we can merge it ?

Thanks

vasilvestre:
> LGTM, except for prefixing the php core functions which imho hurts readability

As it allow you to avoid local namespace override and allow some opcode optimisation, prefixing is important :)

oallain:
Thank you guys for reading and approval :smile:

PR #170, opened 2019-02-22
[WIP] add import/export of products

oallain:
Issue #4 
Import and Export Products datas:

- 'Code'
- 'Name'
- 'Description'
- 'Short_description'
- 'Meta_description'
- 'Meta_keywords'
- 'Main_taxon'
- Attributes (one column by attribute)

TO DO:

- [x] Import - Update product details
- [x] Import - Update product mainTaxon
- [x] Import - Update product attributes
- [x] Import - Create product
- [x] Export - product details
- [x] Export - product mainTaxon
- [x] Export - product attributes (1 column by attribute)
- [x] Behat - test export UI
- [x] Behat - test export CLI
- [x] Behat - test import UI
- [x] Behat - test import CLI
- [x] README - add doc

PR #196, opened 2019-05-31
Feature/import export product in csv

oallain:
Issue #4 
Import and Export Products datas in CSV file:

- Details (Code, Name, Description, Enabled, ...)
- Channels
- Price
- Taxonomy (Main taxon and Product taxon)
- Attributes (one column by attribute type)
- Media (one column by media type)

Include in this PR :
- [x] Import - Create product
- [x] Import - Update product (based on code)
- [x] Export - Product (with datas)
- [x] Behat - test export UI
- [x] Behat - test export CLI
- [x] Behat - test import UI
- [x] Behat - test import CLI
- [x] PhpSpec - tests
- [x] README - add doc


stefandoorn:
It's an huge PR, and going over it I wonder if all files (e.g. I see some handlers that I recognize from before) are required. Can't we maybe move this to be shared between import & export if it's basically a duplicate?

src/Importer/ImporterResultInterface.php
```diff
@@ -35,4 +37,10 @@ public function getFailedRows(): array;
      * @return float The duration (in milliseconds)
      */
     public function getDuration(): float;
+
+    public function setMessage(string $message): void;
```
stefandoorn: I don't think we should add methods here, everyone using it won't like it ;-) I would suggest adding a 'ImportResultLoggerAwareInterface' e.g. that adds these methods.

tests/Behat/Context/ProductsContext.php
```diff
@@ -0,0 +1,158 @@
...
+    {
+        $this->getSession(null)->visit($hp);
+    }
+
+    /**
+     * Checks that response body contains specific text.
+     *
+     * @Then response should contain :text
+     */
+    public function theResponseShouldContain(string $text): void
```
stefandoorn: Let's move public before protected and before private.

src/Service/ImageTypesProviderInterface.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace FriendsOfSylius\SyliusImportExportPlugin\Service;
+
+interface ImageTypesProviderInterface
+{
+    public function getProductImagesCodesList(?bool $prefix = true): array;
```
stefandoorn: What about making the boolean required or something like this:

```
public function getProductImagesCodesList(): array;
public function getProductImagesCodesWithPrefixList(): array;
```

It makes the intent readable from the method name, instead of a boolean that's harder to grasp without looking at the implementing code.

src/Service/ImageTypesProvider.php
```diff
@@ -0,0 +1,49 @@
...
+                } else {
+                    $attrSlug[] = $attr['type'];
+                }
+            }
+        }
+
+        return $attrSlug;
+    }
+
+    public function extractImageTypeFromImport(array $keys): array
```
stefandoorn: This one is not in the interface, I think we shouldn't have it public or it should be in the interface also.

src/Service/ImageTypesProvider.php
```diff
@@ -0,0 +1,49 @@
...
+    {
+        $types = [];
```
stefandoorn: What about using `array_filter` & `array_map` here and return it directly? It's hard to read what's going on, I would first filter it and then map it so it's clear what's happening and what's left of the original input.

src/Service/ImageTypesProvider.php
```diff
@@ -0,0 +1,49 @@
...
+    public function __construct(ProductImageRepositoryInterface $productImageRepository)
+    {
+        $this->productImageRepository = $productImageRepository;
+    }
+
+    public function getProductImagesCodesList(?bool $prefix = true): array
+    {
+        $attrSlug = [];
+        $productImages = $this->productImageRepository->findTypes();
+        foreach ($productImages as $attr) {
```
stefandoorn: I would try to continue/break as soon as possible, so moving on in the code we really do things. It's harder to understand the deeper indented you have to read code.

```suggestion
        foreach ($productImages as $attr) {
            if (empty($attr['type'])) {
                continue;
            }
            
            $attrSlug[] = ($prefix ? self::IMAGES_PREFIX : '') . $attr['type']
        }
```

src/Importer/ImporterResultInterface.php
oallain: Yes, it's better :)

tests/Behat/Context/ProductsContext.php
oallain: Yes, sorry

src/Service/ImageTypesProviderInterface.php
oallain: Yes, i do that quickly, but bad ... i changed it :)

src/Service/ImageTypesProvider.php
oallain: I forget it in interface ;)

src/Service/ImageTypesProvider.php
oallain: II changed it :smile:

src/Service/ImageTypesProvider.php
oallain: Yeah, great :)

oallain:
Hello @stefandoorn 

I started this PR before using it on a project. so she wasn't complete.

She is currently used on a project in Staging, so, I improve it continuously. I have recently added for example : management of images, prices, ...

She does a lot of things, so she is (too) very fat and I'm sure she is difficult to review.

Many code comme from https://github.com/FriendsOfSylius/SyliusImportExportPlugin/pull/170
And routing change is now already merge in PR https://github.com/FriendsOfSylius/SyliusImportExportPlugin/pull/194

oallain:
ping @lsmith77 @stefandoorn @solutionDrive-Alt

lsmith77:
I am on vacation. feel free to finalize and merge the PR without my feedback

stefandoorn:
I still find it hard to review due to the size, but I've tried to cover most of it and it looks good to me. Thanks @oallain!

src/Processor/ProductProcessor.php
```diff
@@ -0,0 +1,383 @@
...
+        $product->setMetaDescription(substr($data['Meta_description'], 0, 255));
+        $product->setMetaKeywords(substr($data['Meta_keywords'], 0, 255));
+        $product->setSlug($product->getSlug() ?: $this->slugGenerator->generate($product->getName()));
+    }
+
+    private function setVariant(ProductInterface $product, array $data): void
+    {
+        $productVariant = $this->getProductVariant($product->getCode());
+        $productVariant->setCurrentLocale($data['Locale']);
+        $productVariant->setCurrentLocale($data['Locale']);
```
vasilvestre: Duplicated line

PR #203, opened 2019-10-10
travis - add Sylius 1.6

oallain:


oallain:
Scrutinizer is always wrong, do you think we have to disable it? @stefandoorn @lsmith77

stefandoorn:
It seems to have issues with the amount of dependencies. Maybe we should exclude some or maybe we can even exclude that whole step? Alternative is to remove Scrutinizer, as we have some other quality tools already.

oallain:
> It seems to have issues with the amount of dependencies. Maybe we should exclude some or maybe we can even exclude that whole step? Alternative is to remove Scrutinizer, as we have some other quality tools already.

I think we can remove scrutinizer too. But I haven't the right now 😔

lsmith77:
fine for me .. BTW as you may have noticed, I am really not able to out time into Sylius let alone this Bundle :-/

Issue #208, opened 2019-11-05
[UI] move import form to a separate page


tautelis:
As I have worked with several ecommerce systems the standard seems to be having import/export buttons in top right corner instead of stuffing all the things on the index page of the resource. Also form errors would be shown in standard symfony/sylius manner (unlike now). Additionally we could add available fields list or something when having a dedicated page for import.

3 thumbs up and I will make a PR myself :)

Issue #71, opened 2018-03-04
Resource name in Import Form is singular


stefandoorn:
https://github.com/FriendsOfSylius/SyliusImportExportPlugin/blob/master/src/Resources/views/Crud/import_form.html.twig#L7

It's now: `Country import`, while `Import countries` would be nicer. Currently the 'country' term is injected, not the plural version of it - countries.

PR #209, opened 2019-11-11
Move import form to separate page

tautelis:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?   | yes in UI, no in CLI
| Fixed tickets | #208 #71

Main goal of this PR:
* move import form to separate import page so the index page is not cluttered

Improvements:
* having import button in top is intuitive when coming from other e-commerce platforms
* most of the import form data validation is moved from `Controller` to `ImportType`
* JS script added to display the name of the selected file for upload
* since this will be a distributable package import route was prefixed with `fos_sylius_import_export_` to avoid clashing with any project level routes prefixed with `app_`
* importer tag types replaced with full sylius resource name same as exporter tags to keep the consistency and avoid sylius resource name ambiguity

**UI import before:**
index-page-before

**UI import after:**
index-page-after

import-page


tautelis:
well, the rule of thumb is if you keep the PR open for too long the conflicts appear... why are we waiting this much? is there any problem with this PR? Things like this demotivates to contribute to open source projects.

PR #215, opened 2019-12-19
Check if json content is valid

FlorentinGarnier:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?   | no
| Fixed tickets | #... 

Throw a new exception if content contain null by json_decode return

### DIFF
--- a/src/Importer/JsonResourceImporter.php
+++ b/src/Importer/JsonResourceImporter.php
@@ -40,7 +40,7 @@ final class JsonResourceImporter extends ResourceImporter implements SingleDataA
 
         $data = json_decode($contents, true);
 
-        if (is_null($data)) {
+        if (null === $data) {
             throw new ImporterException(sprintf('File %s is not a valid json', $fileName));
         }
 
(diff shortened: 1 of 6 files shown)


FlorentinGarnier:
Hello,

Build has faile but it seems its from Sylius

From PHP Stan 
```
274    Call to an undefined method object::setPrice().                         
275    Call to an undefined method object::setOriginalPrice().                 
308    Parameter #1 $channel of method                                         

         Sylius\Component\Channel\Model\ChannelsAwareInterface::addChannel()     

         expects Sylius\Component\Channel\Model\ChannelInterface, object given.  
```

FlorentinGarnier:
Normally it's done

src/Importer/JsonResourceImporter.php
```diff
@@ -38,7 +38,13 @@ public function import(string $fileName): ImporterResultInterface
             throw new ImporterException(sprintf('File %s could not be loaded', $fileName));
         }
 
-        foreach (json_decode($contents, true) as $i => $row) {
+        $data = json_decode($contents, true);
+
+        if (is_null($data)) {
```
oallain: You need to write
```suggestion
        if (null === $data) {
```

for fix Ecs error in Travis.

oallain:
Thanks @FlorentinGarnier 👍

Issue #165, opened 2019-02-04
Export filename ends with '.html' when using web ui


siven-xtn:
When exporting from the web UI, the export filename ends with '.html'. 

Issue found on Sylius 1.3.7

The solution is to add the 'Content-Type' header to the response.

A simple fix (but not the right one) is to add the following after line 80 in the file vendor/friendsofsylius/sylius-import-export-plugin/src/Controller/ExportDataController.php
```
$response->headers->set('Content-Type', 'application/' . $format);
```

alexweggs:
Weirdly, i have this bug now but was working fine til today !

I confirmed if you had the content-type header it works fine.

oallain:
Resolved by #223

PR #223, opened 2020-01-21
Add content type to header

alexweggs:
Add the content type to header for correct file format download

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?   | no
| Fixed tickets | #165  

Fix the bad .html extension when download exported file from Web UI


oallain:
Thanks @alexweggs !

Issue #237, opened 2020-05-27
Support for Symfony 4.4


andychan94:
Plugin does not seem to be compatible with Symfony 4.4.
```
friendsofsylius/sylius-import-export-plugin 0.18.0 requires symfony/stopwatch ^3.3 | ^4.1
```
Symfony 4.4 comes with symfony/stopwatch v5.0.8
Is there a workaround?

andychan94:
I forked the repo and added the version needed to stopwatch and it seems to be working fine
```"symfony/stopwatch": "^3.3 | ^4.1 | ^5.0",```

oallain:
Hi @andychan94 ,

Do you want to make PR here for add it ?

PR #239, opened 2020-06-10
Support for Symfony 4.4

MyUtd4Life:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?   | no
| Fixed tickets | #237 

Update the dependency of the Symfony Stopwatch component (thanks to @andychan94) to add support for Symfony 4.4.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
         "php": "^7.2",
         "sylius/sylius": "^1.4",
         "portphp/portphp": "^1.2",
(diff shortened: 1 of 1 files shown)


oallain:
Hi @MyUtd4Life 

Travis tests are in error, but I don't think it's due to your PR.

So I validate it, nevertheless it would be good that we take the time to fix the tests in another PR :wink:

Issue #241, opened 2020-07-22
Composer installation fails on symfony 4.4


lwillems:
Hello,

Got the following error with composer when trying to install on sylius 1.8.0-dev (SF 4.4.7)

```
  Problem 1
    - Conclusion: don't install symfony/symfony v4.4.7
    - Installation request for symfony/web-profiler-bundle (locked at v4.4.7, required as ^4.4) -> satisfiable by symfony/symfony[v4.4.7], symfony/web-profiler-bundle[v4.4.7].
    - Installation request for symfony/dom-crawler (locked at v5.0.7) -> satisfiable by symfony/dom-crawler[v5.0.7].
    - Installation request for symfony/dotenv (locked at v4.4.7, required as ^4.4) -> satisfiable by symfony/dotenv[v4.4.7], symfony/symfony[v4.4.7].
    - Installation request for symfony/framework-bundle (locked at v4.4.7) -> satisfiable by symfony/framework-bundle[v4.4.7], symfony/symfony[v4.4.7].
    - Installation request for friendsofsylius/sylius-import-export-plugin 0.17.0 -> satisfiable by friendsofsylius/sylius-import-export-plugin[0.17.0].
    - Installation request for symfony/stopwatch (locked at v5.0.7) -> satisfiable by symfony/stopwatch[v5.0.7].
    - symfony/messenger v4.4.7 conflicts with symfony/symfony[v3.4.7].
    - symfony/symfony v3.4.7 conflicts with symfony/messenger[v4.4.7].
    - Conclusion: install symfony/symfony v3.4.7
    - Installation request for symfony/messenger (locked at v4.4.7) -> satisfiable by symfony/messenger[v4.4.7], symfony/symfony[v4.4.7].
```

Any idea to solve this issue ?

s4iya:
Same here, #239 didn't seem to have fixed it for me. For now, the only way I got it to work was by downgrading to 0.16.

```
"require": {
        "friendsofsylius/sylius-import-export-plugin": "^0.16.0",
        "symfony/stopwatch": "^3.3 | ^4.1 | ^5.0"
}
```

oallain:
Hello @lwillems and @s4iya 

Sylius have make a fix who was merged in Sylius 1.7.
See at https://github.com/Sylius/Sylius/pull/11802

It fix you issue ?

PR #251, opened 2021-03-06
Update bundle for Sylius 1.9 / Symfony 4.4+||5.2+

Gogopex:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?   | no
| Fixed tickets | #241, update to Sylius 1.9

I was met with quite a few errors trying to install this plugin on a brand new Sylius 1.9 webapp. I decided to take some time to fix these, in case you guys want to update this bundle for Symfony 1.9

This PR includes a bunch of updates to a few files:
- a bump to the `symfonystopwatch` version, allowing the bundle to install succesfully after being `composer require`'d 
- an update to `src/DependencyInjection/Configuration.php` which ensures that it both works on Symf 5.2, but also maintains BC with 4.1 and older
- ~replaced the `ObjectMapper` typings with `EntityManagerInterface`~
- added `: int`typing and `return 0;` to the `execute()`  of all commands (required by Symf 4.4+. I did not use `return Command::Success` to maximize BC.) 
- updated factory declaration of `sylius.class_metadata.product_image` in `services.yml` to symf4.4+ format

With just these changes, the bundle installs and all features are working fine. 👍 

_____

Note: the builds fail because they are using the old dependencies: 

```
DEPENDENCY_VERSIONS="sylius/sylius:1.6.* symfony/symfony:4.2.*"
```

They should pass once the depency versions are `"sylius/sylius:1.9.* symfony/symfony:5.2.*"` 👍


composer.json
```diff
@@ -1,13 +1,13 @@
 {
-    "name": "friendsofsylius/sylius-import-export-plugin",
+    "name": "mylittleparis/sylius-import-export-plugin",
```
Gogopex, line 2:
```suggestion
    "name": "friendsofsylius/sylius-import-export-plugin",
```

src/Importer/JsonResourceImporter.php
```diff
@@ -4,14 +4,14 @@
...
-use Doctrine\Common\Persistence\ObjectManager;
+use Doctrine\ORM\EntityManagerInterface;
 use FriendsOfSylius\SyliusImportExportPlugin\Exception\ImporterException;
 use FriendsOfSylius\SyliusImportExportPlugin\Processor\ResourceProcessorInterface;
 
 final class JsonResourceImporter extends ResourceImporter implements SingleDataArrayImporterInterface
 {
     public function __construct(
-        ObjectManager $objectManager,
+        EntityManagerInterface $objectManager,
```
lsmith77, line 14: why was this change needed?

lsmith77:
thank you! 

do you also have time to look into the travis setup?

src/Importer/JsonResourceImporter.php
lsmith77: ie. why the switch to the ORM specific interface

Gogopex:
> thank you!
> 
> do you also have time to look into the travis setup?

No problem 👍 I can take a look at the travis setup too, I will update this PR once it's done !

lsmith77:
BTW most OSS projects have migrated to GithubActions due to Travis slowness. but then again maybe due to everyone having moved Travis is now acceptable in speed.

src/Importer/JsonResourceImporter.php
Gogopex: I was encountering an error in both importers when trying to run any export/import commands, something of the sort

```
Fatal Error: 
Argument 1 passed to XXX must be an instance of Doctrine\Common\Persistence\ObjectManager, instance of ContainerNrPkzm2\EntityManager_9a5be93 given
```

The short version is that to circumvent this, I changed these two typehints to reflect the object that was being injected 😄 

The long version: I scratched my head at this for a while as 
- ObjectManager is an interface
- EntityManagerInterface itself implements ObjectManager
But there seemed to be shenanigans I did not understand with what is being injected. `ContainerNrPkzm2\EntityManager_9a5be93` itself seems to implement the wrong thing.

Some source: https://stackoverflow.com/a/45678423/2926899

lsmith77: can you have a look at what `class_implements` returns?

https://github.com/mvpatel/symfony/blob/master/vendor/doctrine/orm/lib/Doctrine/ORM/EntityManagerInterface.php

Gogopex: can do 👍

Gogopex:
Ok, this first attempt with Travis did not go too well. I'm not super familiar with it but I'll attempt to run the pipeline locally if possible to tinker with it till it runs. 👍

lsmith77:
I could imagine using composer 2 could solve the memory issue

src/Importer/JsonResourceImporter.php
Gogopex: ```dump(class_implements($objectManager));```

Class: `ContainerY84tjkD\EntityManager_9a5be93`

```
ResourceImporter.php on line 47:
array:6 [
  "Doctrine\Persistence\ObjectManager" => "Doctrine\Persistence\ObjectManager"
  "Doctrine\ORM\EntityManagerInterface" => "Doctrine\ORM\EntityManagerInterface"
  "ProxyManager\Proxy\VirtualProxyInterface" => "ProxyManager\Proxy\VirtualProxyInterface"
  "ProxyManager\Proxy\ValueHolderInterface" => "ProxyManager\Proxy\ValueHolderInterface"
  "ProxyManager\Proxy\ProxyInterface" => "ProxyManager\Proxy\ProxyInterface"
  "ProxyManager\Proxy\LazyLoadingInterface" => "ProxyManager\Proxy\LazyLoadingInterface"
]
```

It seems I was wrong on this. I swapped back the typing to `ObjectManager` in ResourceImporter.php and it worked with no issues on Sylius 1.9, Symfony 5.2.3. It must have been some cache and sleep deprivation.

I will update the PR reverting this change, and also update the two Import commands I did not update yesterday night. 👍

Gogopex:
I made some progress on having a _passing_ travis job 🙏 

However, the builds are still failing. There is something obvious I must be missing.
Locally I've had no issues with the composer update and everything is working fine, but the Travis job keeps failing on different dependencies mismatches.

```shell
export DEPENDENCY_VERSIONS="sylius/sylius:1.9.* symfony/symfony:5.2.*"
composer require ${DEPENDENCY_VERSIONS} --no-update
composer update --prefer-dist
```

So I'll be using my fork for now, but I'll come back to this Travis job during the week. If anyone wants to give it a go, feel free !

composer.json
```diff
@@ -5,10 +5,10 @@
     "license": "MIT",
     "require": {
         "php": "^7.2",
-        "sylius/sylius": "^1.4",
+        "sylius/sylius": "^1.9",
```
lsmith77, line 8: I am not per-se opposed to this but is it really necessary to bump up to 1.9? would a lower version not work?

composer.json
```diff
@@ -26,23 +26,23 @@
...
+        "friends-of-behat/symfony-extension": "^2.2",
         "friends-of-behat/variadic-extension": "^1.1",
         "lakion/mink-debug-extension": "^1.2.3",
-        "phpspec/phpspec": "^5.0",
-        "phpstan/phpstan-webmozart-assert": "^0.11.1",
-        "phpstan/phpstan-symfony": "^0.11.2",
-        "phpstan/phpstan-doctrine": "^0.11.2",
-        "phpstan/phpstan-shim": "^0.11.4",
+        "phpspec/phpspec": "^7.0",
+        "phpstan/phpstan": "0.12.68",
```
lsmith77, line 33: you locked those version to “fix” the memory limit issue on travis ci?

src/DependencyInjection/Configuration.php
```diff
@@ -43,7 +49,7 @@ public function getConfigTreeBuilder(): TreeBuilder
                     ->end()
                 ->end()
             ->end()
-        ->end();
+            ->end();
```
lsmith77, line 52: I guess this change should be undone

Gogopex, line 52:
```suggestion
        ->end();
```

src/Importer/JsonResourceImporter.php
lsmith77: hmm .. the issue is that this means ODM’s are no longer supported. then again not sure how real of a use case this is in case of Sylius.

I send out a tweet to get more feedback on ways we can make it work without this change potentially
https://twitter.com/lsmith/status/1368645554868224004?s=21

composer.json
Gogopex: Yes, it was one of my attempts to fix it. 

I took inspiration from the composer.json over at sylius/sylius: https://github.com/Sylius/Sylius/blob/master/composer.json#L187 thinking that if it was done over there, it wasn't totally criminal to give it a try here!

Gogopex: I will give this a go this weekend 👍

oallain (commented):
Thanks @Gogopex for this PR :)

Travis is indeed in perpetual error, it's annoying.
To switch to Github Action, you can get inspiration from the PluginSkeleton file
https://github.com/Sylius/PluginSkeleton/blob/1.8/.github/workflows/build.yml

If you don't have time to do it, say it and I will try to do it quickly to validate your PR because she is important 🚀

Gogopex:
Thank you, @oallain ! I decided to give up on fighting with Travis and started exactly what you are suggesting this morning: https://github.com/MyLittleParis/SyliusImportExportPlugin/pull/6/files

Work got in the way but once I get the github action working on the fork, **I will include it in this PR** !

I do have one question: what do you think about dropping the support for Sylius <1.7 in order to reduce the number of builds that would run?

oallain:
> Thank you, @oallain ! I decided to give up on fighting with Travis and started exactly what you are suggesting this morning: https://github.com/MyLittleParis/SyliusImportExportPlugin/pull/6/files
> 
> Work got in the way but once I get the github action working on the fork, **I will include it in this PR** !

So cool 😎

> I do have one question: what do you think about dropping the support for Sylius <1.7 in order to reduce the number of builds that would run?

Yes, to drop Sylius 1.7 and less. But keep 1.8 version.

> Edit: github actions are currently down..! https://www.githubstatus.com/

🤕

lsmith77:
agreed for the bump to 1.8 .. and oh the irony of life with GA being down.

but super appreciative of your commitment and work here @Gogopex

Gogopex:
Hi guys ! I've finally gotten around to finishing this up, here's what's new:

- I've based most of build pipeline on SyliusSkeletonPlugin's as per @oallain suggestion
- I had to update the tests/Application folder. Again, I inspired myself entirely on SyliusSkeletonPlugin
- There are some version changes in the composer.json, which I also took from SyliusSkeletonPlugin

The pipeline is now fully working; however it now crashes on the Phpstan's step, which doesn't seem to have anything to do with this PR 👍 

@oallain @lsmith77 see https://github.com/FriendsOfSylius/SyliusImportExportPlugin/actions/runs/683847071 for where it fails 

There is two solutions ahead imo:
- Someone fixes all of these PHPStan issues
- We simply disable/comment/remove the PHPStan step

I'm not sure I can fix all of these issues this week, but I might be able to in the next 10 days, if you guys are too busy. 👍

lsmith77:
I think we should disable PHPStan in this PR and then have a follow up PR addressing this topic. This PR and you have been burdened so much already without a merge.

Will do a deeper review tomorrow.

oallain (commented):
Hello @Gogopex ,

Yes, just comment the PHPStan step, we will correct it in a future PR.
You've done so much already, I think we can merge like this.

Thanks !

src/Importer/JsonResourceImporter.php
mbabker: You should be able to typehint `Doctrine\Persistence\ObjectManager`, I don't think I've ever seen the deprecation notice before where it specifically calls out to use the ORM's interface TBH.  But, I did just monkey patch an app I'm trying to finish upgrading to Sylius 1.9 with the persistence namespace change and had zero issue using the base `ObjectManager`.

src/DependencyInjection/Configuration.php
```diff
@@ -14,8 +14,14 @@ final class Configuration implements ConfigurationInterface
      */
     public function getConfigTreeBuilder(): TreeBuilder
     {
-        $treeBuilder = new TreeBuilder();
-        $rootNode = $treeBuilder->root('fos_sylius_import_export');
+        if (method_exists(TreeBuilder::class, 'getRootNode')) {
```
mbabker, line 17: Since `getRootNode()` was added in 4.1, and the minimum Symfony version this PR supports is 4.4, you don't need the B/C check here.

oallain: IMHO, we can use :
```suggestion
        $treeBuilder = new TreeBuilder('');
        $rootNode = $treeBuilder->getRootNode('fos_sylius_import_export');
```

Gogopex:
Hello,

I've taken into account @mbabker's suggestion, thanks 👍 

Turns out I also had to comment out PHPSpec [(link to example build)](https://github.com/MyLittleParis/SyliusImportExportPlugin/runs/2187351801?check_suite_focus=true) along with Behat [(link to example build)](https://github.com/MyLittleParis/SyliusImportExportPlugin/runs/2195872178?check_suite_focus=true) for now, otherwise the builds would not pass (the equivalent builds were failing on Jenkins before this PR 😦).

I agree a second PR to fix PHPStan and PHPSpec would be great, along with a PR for the Behat tests! Then everything would be all tidied up!

src/Importer/JsonResourceImporter.php
lsmith77: this remains the only concern I have left over merging this PR .. I have reached out to the Symfony community for advice once more.

lsmith77: do we have a test case that fails if we revert this change? if not, could we add one?

lsmith77: “ You’re using the persistence interface to autowire the entity manager. In future versions of DoctrineBundle (next major release) you’ll have to either alias this yourself or, better yet, inject the manager registry”

https://symfony-devs.slack.com/archives/C3FQPE6LE/p1616851875129800?thread_ts=1616793098.118500&channel=C3FQPE6LE&message_ts=1616851875.129800

Gogopex: I followed the advice you guys gave and reverted the typing and tested the import command and everything worked fine. I still think there is something there: both https://github.com/doctrine/DoctrineBundle/blob/2.2.x/UPGRADE-2.0.md#service-aliases and the fact that I did encounter the error from the original comment is strange 🤔 

It would be great if someone else could give the import command a try and confirm that it does indeed work just fine with the ObjectManager typing 🙏 

tl;dr it works just fine with the ObjectManager typing, as @mbabker implied 👍

src/Importer/ResourceImporter.php
```diff
@@ -38,7 +38,7 @@ class ResourceImporter implements ImporterInterface
 
     public function __construct(
         ReaderFactory $readerFactory,
-        ObjectManager $objectManager,
+        EntityManagerInterface $objectManager,
```
lsmith77, line 41: please also revert here then.

lsmith77:
amazing work!

I have created follow up tickets for PHPStan and PHPSpec

Issue #255, opened 2021-03-27
fix PHPStan


lsmith77:
follow up to https://github.com/FriendsOfSylius/SyliusImportExportPlugin/pull/251

PR #257, opened 2021-03-28
Fix PHPStan and enable it in CI

vasilvestre:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?   | no
| Fixed tickets | https://github.com/FriendsOfSylius/SyliusImportExportPlugin/issues/255

Help wanted

- There's some design issue with type, how should we handle this ?
- phpstan/phpstan-symfony should fix command argument type problems but it do not

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,6 +1,12 @@
 parameters:
+    level: max
+    paths:
+        - src
     reportUnmatchedIgnoredErrors: false
     checkMissingIterableValueType: false
+
+    symfony:
+        container_xml_path: tests/Application/var/cache/dev/Tests_FriendsOfSylius_SyliusImportExportPlugin_Application_KernelDevDebugContainer.xml
 
     excludes_analyse:
         # Makes PHPStan crash
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -6,7 +6,7 @@ parameters:
     checkMissingIterableValueType: false
 
     symfony:
-        container_xml_path: tests/Application/var/cache/dev/Tests_FriendsOfSylius_SyliusImportExportPlugin_Application_KernelDevDebugContainer.xml
+       console_application_loader: src/FOSSyliusImportExportPlugin.php
 
     excludes_analyse:
         # Makes PHPStan crash
--- a/src/Controller/ImportDataController.php
+++ b/src/Controller/ImportDataController.php
@@ -72,7 +72,7 @@ final class ImportDataController
         }
         $referer = $request->headers->get('referer');
 
-        return new RedirectResponse($referer);
+        return new RedirectResponse($referer ?? '');
     }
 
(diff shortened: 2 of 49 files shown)


vasilvestre:
phpstan/phpstan-symfony should deal with Command InputOption and InputArgument as stated here https://github.com/phpstan/phpstan-symfony/issues/20 but I can't get it working :(

src/Controller/ImportDataController.php
```diff
@@ -72,7 +72,7 @@ public function importAction(Request $request): RedirectResponse
         }
         $referer = $request->headers->get('referer');
 
-        return new RedirectResponse($referer);
+        return new RedirectResponse($referer ?? '');
```
vasilvestre, line 75: I'm really not sure about this one

phpstan.neon
```diff
@@ -1,7 +1,13 @@
 parameters:
+    level: max
+    paths:
+        - src
     reportUnmatchedIgnoredErrors: false
     checkMissingIterableValueType: false
 
+    symfony:
+       console_application_loader: src/FOSSyliusImportExportPlugin.php
```
vasilvestre, line 9: @ondrejmirtes do you understand the error ? This seems related to https://github.com/phpstan/phpstan-symfony/issues/46

Fatal error: Uncaught Error: Call to a member function all() on int in /home/ubuntu/projects/SyliusImportExportPlugin/vendor/phpstan/phpstan-symfony/src/Symfony/ConsoleApplicationResolver.php:56

src/Exporter/ResourceExporter.php
```diff
@@ -42,6 +42,12 @@ public function __construct(
      */
     public function setExportFile(string $filename): void
     {
+        if (null === $this->writer) {
```
vasilvestre, line 45: JSONResource Exporter do not require this writer and just didn't called the parent construct (this one)

I modified a bit code and services but I'm not sure this is the best way to solve this.

phpstan.neon
ondrejmirtes: Yes, you're loading this file: https://github.com/FriendsOfSylius/SyliusImportExportPlugin/blob/master/src/FOSSyliusImportExportPlugin.php which isn't a script returning Application instance but just a file with a class that doesn't return anything.

vasilvestre: > Yes, you're loading this file: https://github.com/FriendsOfSylius/SyliusImportExportPlugin/blob/master/src/FOSSyliusImportExportPlugin.php which isn't a script returning Application instance but just a file with a class that doesn't return anything.

In case of library, how should this be handled please ?

ondrejmirtes: You need a file that looks like this:

```php
<?php

$app = new Application();
$app->add(new MyCommand());

return $app;
```

Issue #260, opened 2021-04-20
Exporter relies on deprecated `money_format()` function


mbabker:
The [`money_format()`](https://www.php.net/manual/en/function.money-format.php) function is deprecated in PHP 7.4 and removed in PHP 8.0.  The function is still used in [`FriendsOfSylius\SyliusImportExportPlugin\Exporter\Transformer\Handler\IntegerToMoneyFormatHandler::process()`](https://github.com/FriendsOfSylius/SyliusImportExportPlugin/blob/c61b08bbdd67827e3c6db1f0a4cfbe469737c0ab/src/Exporter/Transformer/Handler/IntegerToMoneyFormatHandler.php#L31).

oallain:
Hello @mbabker 

Do you think [NumberFormatter::formatCurrency](https://www.php.net/manual/fr/numberformatter.formatcurrency.php) is the right choice ?

Isabellebengrine:
I have this problem as well; I just installed the plugin and the command line works but if I click on the 'export' button on the order admin page, I get this error about the money_format... so how can I fix that please ?

Isabellebengrine:
By doing this:
protected function process($key, $value): ?string
    {
        $locale = '';
        if(isset($GLOBALS['request']) && $GLOBALS['request']) {
            $locale = $GLOBALS['request']->getLocale();
       }
        $fmt = new \NumberFormatter( $locale, \NumberFormatter::CURRENCY );
        return $fmt->formatCurrency($value/100, "EUR");
    }

 it works now ...
However, being only a beginner, I would appreciate if someone here could review this code and check if this is the right way to do it... also, what can I do to override this file in my sylius app ?
Thanks in advance!

harunbuljubasic:
Any news here? I have the same problem with PHP 8.1

PR #268, opened 2021-09-03
Updated IntegerToMoneyFormatHandler.php to fix process method

Isabellebengrine:
Replacing money_format function in process method

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?   | no
| Fixed tickets | #260  

Replaced the money_format function used in the process method to fix error when using the exporter button

### DIFF
--- a/src/Exporter/Transformer/Handler/IntegerToMoneyFormatHandler.php
+++ b/src/Exporter/Transformer/Handler/IntegerToMoneyFormatHandler.php
@@ -28,7 +28,12 @@ final class IntegerToMoneyFormatHandler extends Handler
      */
     protected function process($key, $value): ?string
     {
-        return money_format($this->format, $value / 100);
+        $locale = '';
+        if(isset($GLOBALS['request']) && $GLOBALS['request']) {
+            $locale = $GLOBALS['request']->getLocale();
+       }
+        $fmt = new \NumberFormatter( $locale, \NumberFormatter::CURRENCY );
+        return $fmt->formatCurrency($value/100, "EUR");
     }
 
     protected function allows($key, $value): bool


src/Exporter/Transformer/Handler/IntegerToMoneyFormatHandler.php
```diff
@@ -28,7 +28,12 @@ public function __construct(array $allowedKeys, string $format = '%.2n')
      */
     protected function process($key, $value): ?string
     {
-        return money_format($this->format, $value / 100);
+        $locale = '';
```
andriusvo, line 31: use `LocaleContextInterface`

src/Exporter/Transformer/Handler/IntegerToMoneyFormatHandler.php
```diff
@@ -28,7 +28,12 @@ public function __construct(array $allowedKeys, string $format = '%.2n')
...
+        if(isset($GLOBALS['request']) && $GLOBALS['request']) {
+            $locale = $GLOBALS['request']->getLocale();
+       }
+        $fmt = new \NumberFormatter( $locale, \NumberFormatter::CURRENCY );
```
andriusvo, line 35: `$formatter` or `$numberFormatter`

andriusvo: no spaces inside braces

src/Exporter/Transformer/Handler/IntegerToMoneyFormatHandler.php
```diff
@@ -28,7 +28,12 @@ public function __construct(array $allowedKeys, string $format = '%.2n')
...
+        return $fmt->formatCurrency($value/100, "EUR");
```
andriusvo, line 36: `EUR` should not be hardcoded

andriusvo (changes requested):

oallain (changes requested):
IMO, @Isabellebengrine you need to consider @andriusvo feedbacks.

src/Exporter/Transformer/Handler/IntegerToMoneyFormatHandler.php
mbabker: Using `LocaleContextInterface` makes a lot of assumptions.  I just inlined a version of this handler to an app to un-block upgrading to PHP 8 and ended up hardcoding the locale in for now because the one spot this handler gets used in our app (order export) the handler lacks the context needed to get the locale from the order, which would mean that the field would be exported in the request's active locale (and in a multi-lingual store that may not be desired).

src/Exporter/Transformer/Handler/IntegerToMoneyFormatHandler.php
mbabker: Same issue as I noted above, we don't have the context to get the right currency code to use here.

oallain:
Fixed in [0.23.0](https://github.com/FriendsOfSylius/SyliusImportExportPlugin/releases/tag/0.23.0)

Issue #249, opened 2021-03-01
Prefixed resource generates wrong format options list


armellarcier:
In order to get an import form displayed I have type the tag with the full, prefixed resource name.

```
    sylius.importer.my_plugin.my_resource.xlsx:
        class: FriendsOfSylius\SyliusImportExportPlugin\Importer\ResourceImporter
        arguments:
            - "@app.factory.my_resource.spreadsheet_reader"
            - "@my_plugin.manager.my_resource"
            - "@app.processor.my_plugin.my_resource"
            - "@sylius.importer.result"
            - 100
            - false
            - false
        tags:
            - { name: sylius.importer, type: my_plugin.my_resource, format: xlsx }
```

But when it comes to generating the list of format options, in `src/Form/ImportType.php`'s `buildChoices` method, the `importer_type` option is set to the request's resource name, which is not prefixed.

So `ImporterRegistry::buildServiceName($importerType, 'xlsx')` (in my case) returns `"my_resource.xlsx"` which is not registered in the `ImporterRegistry` since the service name is built with the `"my_plugin"` prefix.

I end up with only the `JSON` option.

cdarken:
I confirm this.
The cause seems to be the matching in the ImportDataController, line 49. The 'app' part is lost, the route only has the resource.

PR #272, opened 2021-11-23
Make service config a bit more flexible

cdarken:
Added option to use 'domain' in service config tag so instead of type taking the value "app.foo" you will use "type: foo" and "domain: app"
This fixes the selection of export format in the admin interface.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?   | no
| Fixed tickets | #249

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,7 +157,7 @@ admin overview panel using the event hook system, ie. `admin/tax-categories/`.
 
 #### Notes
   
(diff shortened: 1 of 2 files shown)


oallain (approved):

oallain:
Thanks @cdarken :rocket:

PR #280, opened 2022-10-06
Use webpack, node 14 and test plugin against Sylius 1.12 and PHP 8.1

vasilvestre:


oallain (approved):
Thanks @vasilvestre 🚀

PR #282, opened 2022-10-07
Enable spec

vasilvestre:
I don't know phpspec and don't get the failures.. i will just fix twig deprecations for now :)

### DIFF
--- a/src/Exporter/Transformer/Handler/IntegerToMoneyFormatHandler.php
+++ b/src/Exporter/Transformer/Handler/IntegerToMoneyFormatHandler.php
@@ -28,7 +28,7 @@ final class IntegerToMoneyFormatHandler extends Handler
      */
     protected function process($key, $value): ?string
     {
-        return sprintf($this->format, $value/100);
+        return sprintf($this->format, $value / 100);
     }
 
     protected function allows($key, $value): bool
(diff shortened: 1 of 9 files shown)


vasilvestre:
There's still an error, the same as https://github.com/FriendsOfSylius/SyliusImportExportPlugin/pull/268. May use it to fix last errors :)

src/Exporter/Transformer/Handler/IntegerToMoneyFormatHandler.php
```diff
@@ -5,30 +5,33 @@
...
+        $this->localeContext = $localeContext;
     }
 
     /**
      * {@inheritdoc}
      */
     protected function process($key, $value): ?string
     {
-        return money_format($this->format, $value / 100);
+        return (new MoneyFormatter())->format($value, "EUR", $this->localeContext->getLocaleCode());
```
vasilvestre, line 34: We could use CurrencyContext here maybe ?

oallain: Hello,

I'm not sure if this is the right solution.
It seems to me that the LocaleContext and CurrencyContext are based on the current user who is doing the export.
So, if an order is placed in dollars and the administrator is configured in Euro then the export is wrong.
What do you think? Have you tested this case?

vasilvestre: Wouldn't adding a symbol in admin UI the best option here ? So they can pick which symbol is used for every specific import.

oallain: Yes, or just pretend that the person doing the export knows what they are doing and leave it as you did.

What bothers me is to block everything because of that.
So I validate it as it is.

oallain (approved):

PR #283, opened 2022-10-10
Test build fix for master

vasilvestre:
1.12 changed thing with swiftmailer and symfony/mailer so the build is now broken


oallain (approved):
Thanks @vasilvestre !

PR #284, opened 2022-10-10
Fix twig deprecations

vasilvestre:
Since Twig 2.7 these feature has been deprecated and now can safely be used as minimal dependency is 2.11 for this package.


oallain (approved):
Thanks @vasilvestre .

PR #286, opened 2022-10-20
Add symfony 6 compatibility

vasilvestre:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?   | yes
| Fixed tickets | #... 

Symfony 6 requires PHP 8.0 which can't be used in this package at the moment


dsbe-ak:
Why is this plugin dependent on the API bundle? It should work without using the API bundle, or not?

oallain:
IMHO, we need to remove compatibility with Sylius 1.8 and Sylius 1.9 to drop support of PHP 7.4 and make this update easier.

oallain (approved):
Looks good to me 👍🏼

